PR #8, opened 2022-05-03
This is the first commit of the ZinZen-scheduler rewritten by me @sokorototo

sokorototo:
@tijlleenders It would be nice if you reviewed this PR


README.md
```diff
@@ -1,15 +1,38 @@
-## Welcome friend!
```
tijlleenders, line 1: The existing README.md should not be overridden.
You can add things though.

tijlleenders:
This is a large single commit.
As it's the first OK, but from now on it would be nicer to have multiple smaller commits.

.vscode/launch.json
```diff
@@ -1,83 +0,0 @@
-{
```
tijlleenders, line 1: Not sure if .vscode should be in the commit...

LICENSE.md
```diff
@@ -1,661 +0,0 @@
-                    GNU AFFERO GENERAL PUBLIC LICENSE
```
tijlleenders, line 1: The license should stay.

tijlleenders:
Should the /out folder be included in the commit?

pulkitvyas08 (approved):
This looks fine. Not sure about the tests and outputs, maybe we can setup a github actions for this in the subsequent PRs to see if we're going in the right direction or not

tijlleenders (changes requested):

README.md
sokorototo: On it

.vscode/launch.json
sokorototo: It shouldn't, somehow I didn't include it in the .gitignore

sokorototo:
Since I plan on setting up both Javascript and Rust CI tests, yes the `out` directory stays in the repo

src/console.rs
```diff
@@ -0,0 +1,61 @@
...
+		write_to_ipc(data);
+		console_log(true, data.len());
+	}
+}
+
+/// Log a string to the console
+pub fn log_buf<S: AsRef<[u8]>>(data: S) {
+	let data = data.as_ref();
+
+	unsafe {
```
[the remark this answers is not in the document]
sokorototo: `.wasm` is single-threaded so there can never be any two threads reading or writing from static mutable memory. In Rust, any `static mut` variable is unsafe because of data races, but what are you racing against in a single-threaded application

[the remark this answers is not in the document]
sokorototo: Plus all writing to the IPC is handled via the `write_to_ipc` function, which is safe to execute. The unsafe part of the logging functions is calling the `extern "C" **` functions

sokorototo:
@tijlleenders all previous issues seem to sorted out

tijlleenders (approved):

PR #14, opened 2022-05-27
Fix/deno ci test

tijlleenders:


sokorototo:
The Deno CI tests are already fixed, that is unless there are any other significant changes in this PR that you would like to discuss

Issue #3, opened 2022-01-22
Add scheduler test


tijlleenders:
Add a test to see verify that fixed appointments (with only one slot possible) get scheduled first... and the most flexible one stays out of the way.

 \>>date<< is the current date of the day the test is run.

**Input:**  
```
{
  goalId: 1,
  title : "shopping",  
  duration: 1,
  start: >>date<< 10:00 (ISO 8601),  
  deadline: >>date<< 13:00 (ISO 8601)
},   
{
  goalId: 2,  
  title: "dentist",  
  duration: 1,  
  start: >>date<< 10:00 (ISO 8601),
  deadline: >> date<< 11:00 (ISO 8601)  
},
{  
  goalId: 3,
  title : "exercise",  
  duration: 1,
  start: >>date<< 10:00 (ISO 8601),  
  deadline: >>date<< 18:00 (ISO 8601)
}  
```
**Expected result:**  

Summary version:  
Goal 1 => One task from 11-12  
Goal 2 => One task from 10-11   
Goal 3 => One task from 13-14  

Detailed version:  
```
pub struct Calendar {
    pub max_time_units: usize,
    pub time_unit_qualifier: String,
    pub tasks: Vec<Task>,  //Will have to discuss which Collection type is best suited
    pub slots: Vec<Slot>,  //Will have to discuss which Collection type is best suited
}
```

Where Vec\<Task> =
```
[  
  {
    task_id: x,
    duration_to_schedule: 0,
    duration_scheduled: 1,
    task_status: TaskStatus::Scheduled,
    goal_id: 1,
  },
  {
    task_id: y,
    duration_to_schedule: 0,
    duration_scheduled: 1,
    task_status: TaskStatus::Scheduled,
    goal_id: 2,
  },
  {
    task_id: z,
    duration_to_schedule: 0,
    duration_scheduled: 1,
    task_status: TaskStatus::Scheduled,
    goal_id: 3,
  },
]
```

and Vec\<Slot> =
```
[  
  {
    task_id: x,
    begin: 11,
    end: 2,
  },
  {
    task_id: y,
    begin: 10,
    end: 11,
  },
  {
    task_id: z,
    begin: 13,
    end: 14,
  },

```

tijlleenders:
@sokorototo I apologize for not specifying this more clearly.  
It is not possible to just guess where I want to go with this.  

[...]

PR #15, opened 2022-05-30
This PR fixes and vastly improves the tests, if they don't pass changes will be much easier

sokorototo:
Branch pollution is real, so whenever we finish a PR we delete the branch, this ensures all branches are as close to `main` as possible to prevent excess merge conflicts


sokorototo:
Hmm I don't have write access? Why? We are only two people after all

tijlleenders:
> Hmm I don't have write access? Why? We are only two people after all. Or is it that I just need an approving review first

Yeah I communicated this by chat. Main is protected now. We'll integrate PRs that close issues, like any project. Thisvsetsbup the project for better collaboration.

test/goals.json
```diff
@@ -1,23 +1,23 @@
...
   {
-    "goalId": 1,
+    "id": 1,
     "description": "shopping",
-    "duration": 1,
-    "start": "2022-05-27T10:00:00.000Z",
-    "deadline": "2022-05-27T13:00:00.000Z"
+    "task_duration": 1,
+    "start": "June 1, 2022 12:00:00",
+    "deadline": "June 2, 2022 18:00:00"
```
tijlleenders, line 7: I understand the format change but the date and time don't have to be changed.

ts/api.ts
```diff
@@ -49,32 +71,41 @@ export interface Goal {
...
 }
 
-// An interface that describes a Task
+/**
+ * An interface that describes a Task
+ */
 export interface Task {
 	goal_id: number,
 	start: DateTime,
 	finish: DateTime,
```
tijlleenders, line 80: Shoukd be deadline

tijlleenders (changes requested):
Onnour way!

test/goals.json date and times are not conform test case.

Also some "finish" instances that need renaming to deadline.

tijlleenders:
Good to see the test passing!

The asserts still need to be added to make the test fail.

sokorototo:
Asserts are already in place

tijlleenders:
> Asserts are already in place

Where?
I only see one assert in the test:
image


What I need is assert the expected outcome:
Three tasks scheduled according to the expectation defined in the issue #3 

I'll write it here too:
Goal 1 => One task from 11-12
Goal 2 => One task from 10-11
Goal 3 => One task from 13-14

tijlleenders (approved):
Just merging this to restart implementation of preprocessor and scheduler logic.

PR #16, opened 2022-07-13
Get first test (#3) working

xy2iii:
Using tiji's last commits from December/January, simpler.

Get the test working. Once it works, setup a very simple testing infra that runs `cargo test` so we can know if it breaks.

- [x] Preprocessor
- [ ] Scheduler
- [ ] Add CI/automated testing (will probably just reuse Newton's stuff)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,24 +5,26 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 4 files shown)


xy2iii:
#3

xy2iii:
Apologies for the messy history as this removes Newton's changes. The relevant changes are in `lib.rs` and `preprocessor.rs`.

tijlleenders:
The struct after the preprocessor is as expected.

xy2iii:
After review: we should just use what's already there on main. Will restart from there, should be quick to port over

PR #27, opened 2022-07-19
Integrate wasm with frontend

xy2iii:
The idea is to publish as a NPM package and let the frontend install it via `npm -i zinzen-scheduler`. It can then use the wasm directly.


xy2iii:
This feels a little complicated. Copied setup from https://github.com/prisma/prisma-fmt-wasm

xy2iii:
Ok, this is too complicated. Instead of trying to automate everything directly, let's get a simple manual setup first.

PR #38, opened 2022-08-19
Simplify DateRange iteration.

egithinji:
Iterator can be implemented directly on DateRange, allowing us to not require the use of the DateRangeIterator struct. Implementing Iterator still allows us to use DateRange in For loops. Please note, however that this adjustment mutates the DateRange.


tijlleenders (approved):
image
Less code and improved readability. Nice!

PR #44, opened 2022-08-25
Dev

egithinji:


tijlleenders:
@egithinji
For next time, please use descriptive branch names:
- Feature branches should be like feature/feature-name
- Bug fixes and patch branches should be like patch/patch-name
- Improvement to code branches should be like update/update-name

A branch 'dev' can be used for a dev release branch - but that is actually our main branch.
It gets deployed to Vercel and if OK - then gets deployed to production via another mechanism.

All tests have to pass before we merge into main.

tijlleenders:
You'll need to step through the code to see where this goes wrong:
image

tijlleenders (approved):
Great!
All set for a new test case!

PR #78, opened 2022-09-14
create simple edge case with 3 goals

JacobJax:
Created a simple schedule edge case with `input.json` and `output.json` in the `basic-test-1` folder.
Help me confirm that the output is correct.

### DIFF
--- a/tests/jsons/basic-test-1/input.json
+++ b/tests/jsons/basic-test-1/input.json
@@ -1,0 +1,33 @@
+{
+   "startDate": "2022-01-01T00:00:00",
+   "endDate": "2022-01-02T00:00:00",
+   "goals": [
+      {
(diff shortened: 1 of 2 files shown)


egithinji (changes requested):
Great first test! Just one observation: I think two tasks mistakenly have task id 2.

tijlleenders (changes requested):
Are you sure the logic of the test case is OK?
Nothing has flexibility of 1, the swimming seems to have the greatest flexibility so gets scheduled first, 11-13 is the slot with the least overlap - so I expect that to be the times for swimming...

egithinji (approved):

tijlleenders (approved):
Eric already approved it so it can be merged.

PR #82, opened 2022-09-16
add basic-test-1

tijlleenders:
changed ordering of test expectation


tijlleenders:
Not merging this as the output should be in chronological order.
Made issue #83 to update scheduler.

PR #84, opened 2022-09-16
Implement splitting of tasks

### LINKED ISSUE #80: Handle tasks that can be scheduled if split.
Currently if we have the following tasks:
```
{
        "taskid": 1,
        "goalid": 1,
        "title": "dentist",
        "duration": 1,
        "start": "2022-09-14T00:00:00",
        "deadline": "2022-09-15T00:00:00",
        "after_time": 12,
        "before_time": 14,
        "flexibility": 2,
}
```
and
```
{
        "taskid": 2,
        "goalid": 2,
        "title": "work",
        "duration": 8,
        "start": "2022-09-14T00:00:00",
        "deadline": "2022-09-15T00:00:00"
        "after_time": 8,
        "before_time": 21,
        "flexibility": 6
      },

```
The scheduler currently will attempt to schedule task 2 first because it has a higher flexibility. However because task 1 prevents it from having a contiguous 8 hour block, it will fail to schedule. Which is inconvenient because someone can work from 8 to 12, go for their dentist appointment and then resume working from 14 to 18.

egithinji:
```
{
        "taskid": 1,
        "goalid": 1,
        "title": "dentist",
        "duration": 1,
        "start": "2022-09-14T00:00:00",
        "deadline": "2022-09-15T00:00:00",
        "after_time": 12,
        "before_time": 14,
        "flexibility": 2,
}
{
        "taskid": 2,
        "goalid": 2,
        "title": "work",
        "duration": 8,
        "start": "2022-09-14T00:00:00",
        "deadline": "2022-09-15T00:00:00"
        "after_time": 8,
        "before_time": 21,
        "flexibility": 6
},
```

Suppose we have Task Dentist and Task Work and Task Work has a higher flexibility so is scheduled first. Prior to this PR, both would fail because while scheduling Work, it would conflict with Dentist's slots. And Dentist would fail because it conflicts with Work's slots.

This PR introduces a new step. Work will still fail initially. Then when attempting to schedule Dentist, it will find a conflict with Work. However the scheduler will now notice that Work is already a failed-to-schedule task, and allow Dentist to go ahead and schedule itself even though it's using slots within Work. It will then split Work into two tasks. One with slots before the dentist appointment, and one with slots after the dentist appointment. The durations of the new tasks will add up to the duration of the original task.

These two new tasks are then added to the list of tasks to schedule and the scheduler does another sweep to try and schedule all tasks. It will keep repeating this process until all tasks are scheduled (the assumption is we are working only with goals that can all be scheduled if split).

This is just for a very simple split test case. We need to test for many more scenarios and determine minimum size to allow splitting of a task (current is 3).

Closes #80.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,5 @@
+#[derive(Debug)]
+pub enum Error {
+    NoConfirmedDate(usize),
+    CannotSplit,
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,9 +2,10 @@ use wasm_bindgen::prelude::*;
 
 pub use goal::Goal;
 pub use input::Input;
-pub use output_formatter::{output_formatter, Error, Output};
+pub use output_formatter::{output_formatter, Output};
 pub use time_slice_iterator::Repetition;
 
+mod errors;
 /// API modules
 mod goal;
 pub mod input;
@@ -31,6 +32,7 @@ interface Input {
 // https://rustwasm.github.io/wasm-bindgen/reference/arbitrary-data-with-serde.html
 #[wasm_bindgen]
 pub fn schedule(input: &JsValue) -> Result<JsValue, JsError> {
+    use errors::Error;
     use output_formatter::*;
     use task_generator::task_generator;
     use task_placer::*;
@@ -45,6 +47,9 @@ pub fn schedule(input: &JsValue) -> Result<JsValue, JsError> {
         Err(Error::NoConfirmedDate(id)) => {
             panic!("Error with task:{id}. Tasks passed to output formatter should always have a confirmed_start/deadline.")
         }
+        Err(e) => {
+            panic!("Unexpected error: {:?}", e);
+        }
         Ok(output) => output,
     };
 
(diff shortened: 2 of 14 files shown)


egithinji:
Please find attached explanatory slides.
[Splitting Task Test Case Zinzen (1).pdf](https://github.com/tijlleenders/ZinZen-scheduler/files/9584083/Splitting.Task.Test.Case.Zinzen.1.pdf)

tijlleenders:
I don't follow the logic in the PR comment.
Dentist has flex 2 so work will go first and succeed.

egithinji:
> I don't follow the logic in the PR comment. Dentist has flex 2 so work will go first and succeed.

I thought that apart from tasks with flex 1, the scheduler proceeds from highest flex to lowest flex?

tijlleenders:
The previous algorithm would select the slot with the *least* conflict for work. We could evolve that to schedule per hour if scheduling as a block does not have an option without conflicts. Scheduling per hour would work the same as scheduling per block - but with a 1h block.

tijlleenders:
I'm not sure if there's a difference in the result but switching/skipping tasks and later coming back to them seems more complicated than reducing the block size and retrying.

tijlleenders (approved):
Test case is passing.
We'll need a new issue that follows the logic of scheduling per hour - but that will fail with the current implementation (which doesn't choose the slot with the least number of conflicts but only slots with zero conflicts).

A good one might be two 8 hour tasks that both have the same allowed time slots?

PR #90, opened 2022-09-19
created new splitting tasks

JacobJax:
Looks like the output from the split tasks has a gap between 15:00 and 16:00.

### DIFF
--- a/tests/deno_tests.ts
+++ b/tests/deno_tests.ts
@@ -83,7 +83,7 @@ Deno.test("single day with split tasks", () => {
 });
 
 // test for splitting tasks variant
-Deno.test("single day with split tasks", () => {
+Deno.test("single day with three goals to be split", () => {
   assertEquals(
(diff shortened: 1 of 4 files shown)


egithinji:
> Looks like the output from the split tasks has a gap between 15:00 and 16:00.

Yes sometimes this happens because the algorithm "respects" the aftertime and beforetime of other tasks.

egithinji (approved):

egithinji (commented):
Please add the new test to the integration tests: both deno_tests.ts and rust_tests.rs

egithinji (changes requested):
Please add the new test to the integration tests: both deno_tests.ts and rust_tests.rs

egithinji (changes requested):
Please also add the new split test to the deno tests.

egithinji (dismissed):

egithinji (approved):

PR #96, opened 2022-09-21
Uncomment realistic-schedule test.

egithinji:
This test had been commented out while #83 was being worked on. Now can be uncommented.


JacobJax (approved):
closes #96

PR #100, opened 2022-09-22
rename test folders

### LINKED ISSUE #97: Simplify and make consistent naming for tests
For example in `rust_tests.rs` we have a test called `new_split_simple_variant` which corresponds to the test directory `splitting-tasks-simple-1`. The naming should be consistent in both the test directories and the tests themselves.

For the split-related tasks, I'd suggest split-1, split-2, etc for the variants (starting with split-1).

JacobJax:
closes #97 

Each of the test folders' name ends with a number starting with `-1` incase new variants of the edge cases will be made in future.


egithinji (changes requested):
Hi could you also adjust the names of the functions to be consistent with the names of the directories? For example instead of `split_tasks_variant`, it should be `split_2_works`

egithinji (changes requested):
The error is a file-not-found from `rust_tests.rs` on line 13. I think you need to rename the directory names in that file to match the new directory names.

egithinji (approved):

PR #108, opened 2022-09-23
Weekends repetition

### LINKED ISSUE #106: Add test for "weekends" repetition
Create a single goal that repeats every weekend between 2022-09-01T00:00:00 and 2022-10-01T00:00:00

JacobJax:
closes #106

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 Cargo.lock
 .vscode
 .cargo
+outputformat.txt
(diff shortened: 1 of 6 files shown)


egithinji (approved):
There were some errors in the output.json (missed one weekend and to put taskids) and rust_tests.rs. There was also a bug in the scheduler. These have been fixed and now passing all tests.

PR #119, opened 2022-09-28
Add build for front end to CI

egithinji:


tijlleenders:
So I see the CI builds the files - but where does the build get added to the repo?
- or is this not possible?

egithinji:
For now I'm adding the build to the repo by including the `/pkg` directory when pushing.
I'm not sure if the CI can add the build to the repo, because wouldn't this trigger another pr event, leading to a cycle?

egithinji:
I was wondering if we could pursue the following options to ensure the front-end repo gets latest build?: 1) Use a github action to push to that repo once the build has completed 2) Publish to npm, and then front end can have it as a dependency.

tijlleenders:
@egithinji I'm ok with either way as long as we automate the process and respect the security boundaries. 
Not sure about giving the repo permission to publish to npm. I haven't searched but there must be some kind of way to park build artefacts on GH after a merge.

tijlleenders:
... git submodules is not the way. 
We've investigated that and decided it is not worth the extra complexity.

egithinji:
Okay will look into it further.

egithinji:
The archiving (uploading) of the build seems to be working. Next I'm working on a script that can be used from front-end to download this build if a new version is available.

tijlleenders:
Can you post general questions in the comments? Nearly missed the whole story in the last commit. I don't check the commit messages systematically.

tijlleenders:
First step would be to run the script locally and create a PR to bumb the version.

I think the script should be in the ZinZen repo, not here - as that's where it will be used.

Don't think GH will allow access to build artifacts without auth or a token : https://github.com/actions/upload-artifact/issues/51 
Guess it leads to abuse?

egithinji:
Copy pasting the last commit message here for easier reference:

This script checks whether the ZinZen-Scheduler repo has a new version
of the wasm build (the contents of `pkg` directory), and if so, downloads the files
and replaces the local `pkg` directory with the newly downloaded files.

If the script is run on the front-end before tests are run, it will ensure the front-end has the latest binary of the scheduler in the `pkg` directory before the tests are run.

It should be run from the same location that the current `pkg` directory resides.

One side effect: it uses a file `artifact_id.txt` to store a unique id
for the last downloaded artifact.

Before running it, the environment needs a `ZINZENTOKEN` environmental variable holding a Github
access token that has read access to the repo:

```
export ZINZENTOKEN={Your github token}
```

Question: Would this be run locally while working on the front end or as part of the front end's CI?

egithinji:
> Don't think GH will allow access to build artifacts without auth or a token : [actions/upload-artifact#51](https://github.com/actions/upload-artifact/issues/51) Guess it leads to abuse?

Yes I guess so.
I want to now include automatically creating a PR to the script so that when a new build is found, it gets added to ZinZen main.
I'll continue this on ZinZen repo since that's where the script should reside.

tijlleenders (approved):

PR #121, opened 2022-09-29
create .txt file inputs

### LINKED ISSUE #120: goalParser inputs
Create `.txt` files for inputs parsed by goalParser

JacobJax:
closes #120


tijlleenders (changes requested):
Please check again.
There are some sloppy mistakes (spaces and double start words).

Also, the input has to be what you would initially type, so 8h instead of 8hours

Tushar-4781 (commented):
If these tests are written for frontend then where is the ouput of these inputs i.e. tags.

Issue #130, opened 2022-10-06
CI should not upload new artifact if the wasm build hasn't changed


egithinji:
Currently a new wasm build will be generated and uploaded as an artifact every time a pr is submitted.
This isn't necessary if for example the pr just included new json tests.

PR #142, opened 2022-10-10
Improve CI

egithinji:
Related to #130.
Set a path filter so that action only runs if there are changes in `src` directory. I'm still researching to confirm if only files in `src` will affect the artifacts generated in `pkg`.


egithinji:
Also I'm suggesting it only be triggered by the PR to main, not by both a push to main and then again by PR to main. i.e. the artifacts will only be uploaded once. @tijlleenders what do you think?

egithinji:
Closing as resolved by #143

PR #153, opened 2022-10-17
Refactor to include Slot struct

### LINKED ISSUE #152: Refactor to include a Slot struct
While working on #150 , I realized there's a need for some refactoring. To start with, we can benefit from a `Slot` struct that will encapsulate some functionality relating to working with time slots.  Currently we're working directly with tuples of `(NaiveDateTime,NaiveDateTime)` and this is becoming trickier as we add more complex features to the scheduler.

The `Slot` struct will at its core represent a contiguous block of time, defined by a start NaiveDateTime, an end NaiveDateTime plus useful methods for working with time slots.

egithinji:
This closes #152 . More refactoring will follow this but will open separate issues for those other changes.

### DIFF
--- a/src/slot.rs
+++ b/src/slot.rs
@@ -1,0 +1,105 @@
+use chrono::{NaiveDateTime, Timelike};
+use serde::{Deserialize, Serialize};
+use std::ops::{Add, Sub};
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy, Deserialize)]
+pub struct Slot {
+    pub start: NaiveDateTime,
+    pub end: NaiveDateTime,
+}
+
+impl Sub for Slot {
+    type Output = Vec<Slot>;
+
+    fn sub(self, other: Self) -> Self::Output {
+        let mut result = Vec::new();
+        if (other.start < self.start) && (other.end <= self.start) {
+            //other is completely before self
+            result.push(self);
+            return result;
+        } else if (other.start >= self.end) && (other.end > self.end) {
+            //other is completely after self
+            result.push(self);
+            return result;
+        } else if (other.start == self.start) && (other.end < self.end) {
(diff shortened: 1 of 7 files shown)


src/slot.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub start: NaiveDateTime,
+    pub end: NaiveDateTime,
+}
+
+impl Sub for Slot {
+    type Output = Vec<Slot>;
+
+    fn sub(self, other: Self) -> Self::Output {
+        let mut result = Vec::new();
+        if (other.start < self.start) && (other.end <= self.start) {
```
tijlleenders, line 16: I would expect six different cases:
- entirely before
- entirely after
- overlapping begin
- overlapping end
- overlapping begin + end (equal or bigger)
- overlapping within (smaller)

Mind the equals for the slot [begin:end[:
- begin is inclusive
- end is exclusive

tijlleenders (changes requested):

src/slot.rs
tijlleenders: Subtraction - I count seven if blocks.
Didn't check details.

egithinji: I've just double-checked and I believe all cases are handled. You can verify from the comments in the `if` blocks or from the unit tests from line 377.

tijlleenders: If we can simplify to 6 if blocks instead of 8 I think we should.

tijlleenders:
Something to think about after this PR: a Task has a Vec<Slot>. This vec can hold none, one, or many Slots. Slot arithmetic should probably take a Vec and output a Vec - or it will lead to inconsistencies which make chaining impossible.

tijlleenders (approved):

PR #156, opened 2022-10-28
Refactor contiguous slots

### LINKED ISSUE #155: Refactor for tasks to store slots of multiple contiguous hours
With the completion of #152 , it's now possible to carry out this refactor.
Currently each task stores a vector of `Slot`s, but each `Slot` is 1hr long and this leads to some complications when dealing with more tricky scenarios. Refactor so that a task stores a vector of `Slot`s where each slot is a contiguous block of several hrs.

e.g.:
- a task with `after_time` 10 and `before_time` 14 on a specific day, should have one `Slot` in it's vec with start 10 and end 14.
- a task that is to be accomplished once between Monday and Friday but between `after_time` 10 and `before_time` 14 will have 5 `Slot`s in it's vec, each `Slot` having start 10 and end 14 (but for different days).

egithinji:
Closes #155 .

### DIFF
--- a/src/slot.rs
+++ b/src/slot.rs
@@ -1,3 +1,4 @@
+use chrono::Duration;
 use chrono::{NaiveDateTime, Timelike};
 use serde::{Deserialize, Serialize};
 use std::ops::{Add, Sub};
@@ -101,5 +102,17 @@ impl Add for Slot {
             //slots that are adjacent to each other
             return self;
         }
+    }
+}
+
+impl Slot {
+    pub fn num_hours(&self) -> usize {
+        let mut hours = 0;
+        let mut time = self.start;
+        while time < self.end {
+            time += Duration::hours(1);
+            hours += 1;
+        }
+        return hours;
     }
 }
--- a/src/task.rs
+++ b/src/task.rs
@@ -1,6 +1,7 @@
 use crate::errors::Error;
 use crate::goal::Goal;
 use crate::slot::Slot;
+use chrono::Duration;
 use chrono::{NaiveDate, NaiveDateTime, Timelike};
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
@@ -22,7 +23,7 @@ pub struct Task {
     pub slots: Vec<Slot>,
     pub confirmed_start: Option<NaiveDateTime>,
     pub confirmed_deadline: Option<NaiveDateTime>,
-    pub internal_index: usize,
+    pub internal_marker: NaiveDateTime,
 }
 
 impl Ord for Task {
@@ -39,6 +40,9 @@ impl PartialOrd for Task {
 
 impl Task {
     pub fn new(id: usize, start: NaiveDateTime, deadline: NaiveDateTime, goal: &Goal) -> Self {
+        //set internal_marker to first possible hour for the task
+        let mut internal_marker = start;
+        internal_marker += Duration::hours(goal.after_time.unwrap_or(0) as i64);
         Self {
             id,
             goal_id: goal.id,
@@ -53,14 +57,17 @@ impl Task {
             slots: Vec::new(),
             confirmed_start: None,
             confirmed_deadline: None,
-            internal_index: 0,
+            internal_marker,
         }
     }
 
     //TODO: The current way this is done may not be entirely accurate for tasks that can be done on
     //multiple days within certain time bounds.
     pub fn calculate_flexibility(&mut self) {
-        let hours_available = self.slots.len();
+        let mut hours_available = 0;
+        for slot in &self.slots {
+            hours_available += slot.num_hours();
+        }
         self.flexibility = hours_available - self.duration + 1;
     }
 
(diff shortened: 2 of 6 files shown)


src/slot.rs
```diff
@@ -103,3 +104,15 @@ impl Add for Slot {
...
+
+impl Slot {
+    pub fn num_hours(&self) -> usize {
+        let mut hours = 0;
+        let mut time = self.start;
+        while time < self.end {
+            time += Duration::hours(1);
+            hours += 1;
+        }
+        return hours;
```
tijlleenders, line 116: Why loop instead of end-start?

src/task.rs
```diff
@@ -53,14 +57,17 @@ impl Task {
             slots: Vec::new(),
             confirmed_start: None,
             confirmed_deadline: None,
-            internal_index: 0,
+            internal_marker,
         }
     }
 
     //TODO: The current way this is done may not be entirely accurate for tasks that can be done on
     //multiple days within certain time bounds.
```
tijlleenders, line 65: Shouldn't all constraints be applied to the Slot vec at Task initialization - leaving only the actually valid Slots?

src/task.rs
```diff
@@ -53,14 +57,17 @@ impl Task {
...
     pub fn calculate_flexibility(&mut self) {
-        let hours_available = self.slots.len();
+        let mut hours_available = 0;
+        for slot in &self.slots {
+            hours_available += slot.num_hours();
+        }
         self.flexibility = hours_available - self.duration + 1;
```
tijlleenders, line 71: This assumes the Task is splittable into 1h portions that can be scheduled independently. Not always true.

Also, is it possible for this to become negative?

src/unit_tests.rs
```diff
@@ -548,3 +399,132 @@ fn adding_slots_to_each_other_works() {
...
+
+fn get_test_tasks_2() -> Vec<Task> {
+    vec![
+        Task {
+            id: 1,
+            goal_id: 2,
+            title: "clean car".to_string(),
+            duration: 2,
+            status: TaskStatus::UNSCHEDULED,
+            flexibility: 0,
```
tijlleenders, line 411: Guess it is ok to initialize this with 0 if flex gets recalculated before any action happens.

tijlleenders (approved):

tijlleenders:
Nice to see this coming along!
Left some comments.

src/task.rs
tijlleenders: I mean that when a Task gets constructed it sgould not be allowed to be in an invalid state. Initializing it with Task::new before knowing which vec<Slot> correspond to its time constraints will make this Task's state invalid.

src/slot.rs
egithinji: Noted and agreed. Had overlooked the `.num_hours()` method available to the `Duration` struct. Will adjust in a new pr

src/task.rs
egithinji: Just to help me understand which part of the process you're referring to, let me recap how it's currently working:

1) The task generator generates tasks from goals with correct constraints based on the goal. i.e. the tasks are given correct start,deadline, and aftertime,beforetime values. These tasks initially have no slots assigned though.
2) Each task is assigned slots based on it's constraints.
3) Each task is scheduled.

Are you referring to what happens in step 1?

src/task.rs
egithinji: Yes currently the code assumes this. Will need to add support for other scenarios.

Regards the negative flexibility, this could happen in the event that the calendar end cuts off the required number of hours to be assigned a task. For example if the calendar ends on a particular day at midnight, and a task `sleep` was to start at 22 and continue onto the next day.

So how I'm currently handling this, is the task's duration will be truncated (preventing the negative from occurring). In the output the sleep would go from 22 to midnight and get scheduled. I'm not sure if this is the best way to handle this and open to suggestions.

src/unit_tests.rs
egithinji: Yes it does. The alternative is to use an `Option` but this makes it unnecessarily verbose.

src/task.rs
tijlleenders: Thanks for the breakdown.
That makes communucation easy.

What I'm referring to is making 2 part of 1. 

Allowing the Tasks to be created in 1 and then 'assigning' (the right word for this case) slots would be OK if you create an extra Task status 'NOSLOTASSIGNEDYET' - but it's easier to not allow Tasks with an inconsistent internal state.

egithinji: Okay I see now. Will consider this after resolving next couple of issues. After completing #159

src/task.rs
tijlleenders: Yeah the Task should have all the info to calculate its flexibility.

The sleep case is interesting.

For now I would prefer just adding a day to the calendar and truncating any Task that starts on the last day - rather than artificially reducing the duration. 

The idea of the Calendar is not that time 'ends' on that datetime (it's not the apocalypse) just rather the end of the requested time range on the calendar. Any Task that starts after the requested range can be discarded. Any Task with a due date before the end of the Calendar (view) must be possible.

src/unit_tests.rs
tijlleenders: If we include setting the correct flexibility after the Slot assignment/generation at Task initialization time, we can make it responsible for always having the correct flexibility at hand.

It can then be done on initialization + after every edit to its slots. Accessing it would then just be a getter to the internal value (like a cache) - not a new calculation every time.

PR #164, opened 2022-10-30
put target back in .gitignore

tijlleenders:
target was not ignored ... probably done when trying to create a link to the build?

Can we put it back so I don't get 600+ commits when running cargo test?


egithinji (approved):
Oh yeah this was removed while working on CI changes a few weeks back

PR #165, opened 2022-10-30
Bug/UI test case without start or deadline

tijlleenders:
@egithinji Can you check this failing testcase?

In principle I see no reason why we can't omit start and/or deadline from a logical point of view.
Is the deserializer not creating a None for the optional values of goal - or is there something else going on?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Cargo.lock
 .vscode
 .cargo
 outputformat.txt
+target/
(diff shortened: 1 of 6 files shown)


egithinji (approved):
There was a bug, now fixed. Thanks for catching it!

tijlleenders:
> There was a bug, now fixed. Thanks for catching it!

@Tushar-4781 Caught it :)

Anyway, @egithinji I ran the update_artifacts.sh script from ZinZen but it is not producing a PR with a new version of the scheduler (https://github.com/tijlleenders/ZinZen/pull/729). Can you check? Did I have to do something on the ZinZen-scheduler repo side?

tijlleenders:
@egithinji I think I figured it out (GH token missing).
Will update the script to check for it (and the other prerequisites - and give an nice error message).

Issue #159, opened 2022-10-29
Refactor how slots are assigned - reduce a monolithic slot by iterating constraints


egithinji:
Implement approach described here:

_Originally posted by @tijlleenders in https://github.com/tijlleenders/ZinZen-scheduler/issues/157#issuecomment-1295746088_

Issue #158, opened 2022-10-29
Refactor `internal_marker`


egithinji:
Some code smell here in the way we keep track of start/deadline combinations to attempt:
https://github.com/tijlleenders/ZinZen-scheduler/blob/502383b38b35d192a3261528f4b08123271d46c7/src/task.rs#L118

egithinji:
Closed via #168

PR #167, opened 2022-11-01
Clean/assign slots

### LINKED ISSUE #162: Handle slot assignment during task initialization
_Originally posted by @tijlleenders in https://github.com/tijlleenders/ZinZen-scheduler/pull/156#discussion_r1008708732_
      
Refactor so that slots are assigned to tasks during task initialization. Consider renaming `slot_assigner` to `slot_generator`.

egithinji:
This closes #162, assigning slots upon initialization of the task via a `slot_generator` module.
It also includes:
- changes to ensure tasks are initialized with their actual flexibility (instead of flexibility of 0).
- a change to how a task that extends beyond the calendar end is treated (e.g. sleep on the last day of calendar). It allows this task to extend to the next day instead of truncating its duration.

Should be easier to work on #158 and #159 after this.


egithinji:
As per your suggestion, adjusted to check whether before and after time are none before looping through slot_generator logic, in order to improve performance.

egithinji:
Have added private calculate_flexibility method and public get_flex method, and also refactored such that slots are removed from tasks by calling the task's .remove_slot() method. 

Not yet calculating flexibility once this is done though, this will require a bit of work so will do it in separate pr.

tijlleenders (approved):
Nice going!

PR #168, opened 2022-11-04
Clean/internal marker

egithinji:
This is related to #158. It improves the code by not allowing the `internal_marker` to be modified externally.

I'd like to attempt an alternate approach that will allow us to avoid using the `internal_marker` altogether: implementing `Iterator` on `Task` for iterating over valid start/deadline combinations for the task. Might be more idiomatic this way.

Marking as draft while I attempt this and compare.

### DIFF
--- a/src/task.rs
+++ b/src/task.rs
@@ -23,7 +23,7 @@ pub struct Task {
     pub slots: Vec<Slot>,
     pub confirmed_start: Option<NaiveDateTime>,
     pub confirmed_deadline: Option<NaiveDateTime>,
-    pub internal_marker: NaiveDateTime,
+    internal_marker: NaiveDateTime,
 }
 
 impl Ord for Task {
@@ -48,8 +48,7 @@ impl Task {
         goal: &Goal,
     ) -> Self {
         //set internal_marker to first possible hour for the task
-        let mut internal_marker = start;
-        internal_marker += Duration::hours(goal.after_time.unwrap_or(0) as i64);
+        let internal_marker = slots[0].start;
         Self {
             id,
             goal_id: goal.id,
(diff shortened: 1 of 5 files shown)


tijlleenders:
Agree that an iterator used by ```task_placer``` is a better approach.

egithinji:
Eliminated need for `internal_marker` variable by creating an iterator. Also made changes in `task_placer` to use `Slot`s wherever possible instead of `(NaiveDateTime,NaiveDateTime)` tuples.
There was also a bug I found in task_placer that would have bitten us in the butt in future in the way it was checking for conflicts between tasks that's been fixed.

src/task.rs
```diff
@@ -38,6 +37,56 @@ impl PartialOrd for Task {
...
+            }
+
+            if index != self.slots.len() - 1 {
+                self.marker = self.slots[index + 1].start;
+            }
+        }
+
+        return None;
+    }
+}
```
tijlleenders, line 88:
```suggestion
pub struct StartDeadlineIterator {
    slots: Vec<Slot>,
    duration: usize,
    marker: NaiveDateTime,
    slot_index: usize,
}

impl StartDeadlineIterator {
    fn new(slots: Vec<Slot>, duration: usize) -> StartDeadlineIterator {
        let marker = slots[0].start;
        let slot_index = 0 as usize;
        StartDeadlineIterator {
            slots,
            duration,
            marker,
            slot_index,
        }
    }
}

impl Iterator for StartDeadlineIterator {
    type Item = Slot;
    fn next(&mut self) -> Option<Self::Item> {
        if self.marker > self.slots[self.slot_index].end - Duration::hours(self.duration as i64) {
            if self.slot_index != self.slots.len() - 1 {
                self.marker = self.slots[self.slot_index + 1].start;
                self.slot_index += 1;
            } else {
                return None;
            }
        }
        let start = self.marker;
        let end = self.marker + Duration::hours(self.duration as i64);
        self.marker += Duration::hours(1);
        return Some(Slot { start, end });
    }
}
```

I found the iteration a bit hard to read.
Whenever I see ```if !(something)``` I like to refactor to the positive form.
That didn't solve the readability.

I suggest we not only keep a marker to the current NaiveDateTime but also a marker to the slot.
This way we don't have to loop through all the slots at every ```next``` and it's more readable?

tijlleenders (approved):

Issue #170, opened 2022-11-06
Improve readability of startdeadline iterator


egithinji:
Add `slot_index` variable.
Write test.

tijlleenders:
Closed by #171

PR #171, opened 2022-11-07
Improve readability of startdeadline iterator

egithinji:
Related to #170 .


egithinji:
Now that I've added a failing test, I'm going to first attempt the suggestion we discussed as-is...

egithinji:
The test fails because of not checking if each slot has enough time for the task's duration.
I'll retain the usage of `slot_index` but will fix this bug.

tijlleenders (approved):
Nice catch!

PR #172, opened 2022-11-09
Clean/custom sort

### LINKED ISSUE #166: task_placer not following algorithm
@egithinji 
The task_placer currently loops through all tasks once to check if any have flexibility 1.
```
    for index in 0..tasks.len() {
        if tasks[index].flexibility == 1 {
            let my_slots = tasks[index].get_slots();
            tasks[index].set_confirmed_start(my_slots[0].start);
            tasks[index].set_confirmed_deadline(my_slots[0].end);
            tasks[index].status = SCHEDULED;
            //slide 10 (remove the assigned slot from other tasks' slot lists)
            remove_slots_from_tasks(&mut tasks, my_slots[0].start, my_slots[0].end);
        }
    }
```

However, this needs to be done after every time a Task is scheduled (TaskStatus::SCHEDULED).

I would suggest implementing a loop (with a loop protection/limit) that goes through all tasks to find an UNSCHEDULED task with flex 1.  
- If found and scheduled (including recalculating all flexibilities impacted) - restart the loop.
- If none UNSCHEDULED with flex 1 are found - pick the one with highest flex, schedule (including recalculating all flexibilities impacten) - restart the loop.
- If no UNSCHEDULED are left (or if loop limit is reached) - exit the loop

egithinji:
Closes #166.
Implemented custom sorting for collections of Tasks by providing custom implementation of `Ord`. It results in tasks being sorted in the following way:
- Tasks with flex 1 first
- Followed by task with highest flex
- Followed by rest of tasks in descending flex order.

This simplified `task_placer` significantly, allowing the bug to be fixed and algorithm better adhered to.


src/goal.rs
```diff
@@ -97,11 +102,10 @@ impl Goal {
             //assign slots that are within the specified after_time and before_time
             let slots = slot_generator(self.after_time, self.before_time, &time_period);
             //calculate flexibility
-            let mut hours_available = 0;
+            let mut flexibility = 0;
             for slot in &slots {
-                hours_available += slot.num_hours();
+                flexibility += slot.num_hours() - self.duration + 1;
```
tijlleenders, line 107:
```suggestion
                flexibility += std::cmp::max(0, slot.num_hours() - self.duration + 1);
```

src/task.rs
```diff
@@ -110,17 +127,15 @@ impl Task {
...
     }
 
-    //TODO: The current way this is done may not be entirely accurate for tasks that can be done on
-    //multiple days within certain time bounds.
     fn calculate_flexibility(&mut self) {
-        let mut hours_available = 0;
+        let mut flexibility = 0;
         for slot in &self.slots {
-            hours_available += slot.num_hours();
+            flexibility += slot.num_hours() - self.duration + 1;
```
tijlleenders, line 133:
```suggestion
            flexibility += std::cmp::max(0, slot.num_hours() - self.duration + 1);
```

tijlleenders (approved):
Nice!

src/goal.rs
egithinji: The `slot_generator` only assigns valid slots for the task so I don't think `slot.num_hours() - self.duration + 1` will be less than 0. If it doesn't wouldn't we rather that the code to panics?

src/task.rs
egithinji: `calculate_flexibility` is called after `remove_invalid_slots` has been called so `slot.num_hours() - self.duration + 1` should never be less than 0. If it is, wouldn't we rather be alerted to this by a `panic`?

src/goal.rs
tijlleenders: You're absolutely right!

src/task.rs
tijlleenders: Yup, agree!

PR #173, opened 2022-11-10
Feat/x per week

egithinji:
Related to #150 . I adjusted the tests to ensure that x per week tasks are always possible within the given calendar bounds. Handling situations where this isn't possible can be done in a separate issue.

The `Repetition` enum has been moved into its own file and is responsible for parsing expressions like "x/week". Then the goal struct uses the `x` to generate `x` tasks for a given time period as it iterates the time periods.

For e.g. if we're given "4/week", we'll iterate through each week, however for each week we'll generate 4 tasks.
Nice thing is with this setup, implementing "x/day", "x/month" etc should be straightforward.

I'm marking as draft for welcoming feedback, but it's ready to close the issue.


tijlleenders (approved):
Nice work!

PR #177, opened 2022-11-11
Implement x per day

### LINKED ISSUE #176: Implement x-per-day functionality

egithinji:
Closes #176 
I had forgotten that the output formatter merges adjacent tasks of the same goal so I had to adjust the test output.jsons in line with this. Also for the same reason added one more test to test behavior when the task is more than 1hr long and can't fit in some positions during the day.


tijlleenders (approved):
Nice!

You want to keep the dbg! in?

egithinji:
> Nice!
> 
> You want to keep the dbg! in?

Yeah I keep putting it back at that place while debugging so I'll keep it for some time as I implement the remaining issues then get rid of it.

Issue #201, opened 2022-11-30
Regression on last scheduler release


tijlleenders:
![image](https://user-images.githubusercontent.com/44887755/204877248-69723306-58a4-4640-901a-8fe493bf9a6c.png)

tijlleenders:
@egithinji Can you fix this before continuing the work on new features?

egithinji:
How come `start` and `deadline` for the goals are not date bounds (the time portion should be 00:00:00)?

tijlleenders:
I guess we found the problem ;)

The scheduler should treat start and deadline as datetime. 

This is different from after/before in that it is not a daily thing - just the extremities constraint of the slot timelime. It can very well be combined with after/before or aome other conatraints.

tijlleenders:
https://github.com/tijlleenders/ZinZen-scheduler/issues/163

egithinji:
Alright will work on this next.

egithinji:
Closed by #203

PR #203, opened 2022-12-02
Fix bug - start and deadline should be datetime not just midnight

egithinji:
Related to #201

### DIFF
--- a/src/goal.rs
+++ b/src/goal.rs
@@ -1,3 +1,4 @@
+use crate::slot::Slot;
 use crate::slot_generator::slot_generator;
 use crate::task::Task;
 use crate::time_slot_iterator::TimeSlotIterator;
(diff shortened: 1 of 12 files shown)


egithinji:
Closes #201.
Also adjusted how certain tasks that go beyond the calendar end/deadline are handled. In my opinion, its a good improvement:

**Previous**
If we had a goal `backup data every 6 hrs duration 4 after_time 15` and the calendar went from say 2022-11-19T00:00:00 to 2022-11-20T00:00:00, the scheduler would schedule two tasks, including the last one which would go past the calendar end/deadline.

**Now**
The first task will be scheduled, and the second one will be included in the `impossible` tasks. The "first_conflict_with" string will indicate, not a goalid, but the string "Passes Calendar End" (actually I'll adjust this to "Passes Deadline/Calendar End").

**Why it's better this way**
It makes no assumptions about the user's situation. For sleep, it might make sense to exceed the deadline/calendar but for some critical tasks which must complete before a deadline, the user might prefer being informed that the task cannot be scheduled, allowing them to either extend the deadline or adjust the goal.

Pinging @Tushar-4781 also because this is relevant to the front end especially the "Passes Deadline/Calendar End" string.

tijlleenders:
Failing if it passes the deadline of the Goal I agree - but it ajoukdn't fail if it only passes the calendar end - since the world will not end ; only the schedule we want to show.

egithinji:
I've added a test for scenario where one task goes past the deadline and reverted the other tests to allow tasks to extend past calendar end when no deadline specified.
Next will refactor as per your comment to make the tests pass.

tijlleenders (approved):
Great!

PR #218, opened 2023-01-03
Feat/options

### LINKED ISSUE #154: Postpone options
Once the calendar is calculated - per Task calculate if there are 0,1,2,3 or more options to reschedule each Task given the other scheduled Tasks don't change - and what is the datetime of the first option to reschedule.

This allows to see how urgent it is to do the Task on the scheduled datetime.

egithinji:
~~Test for specifying behavior described in #154 . I'll implement once this output format is approved.~~
Closes #154


tijlleenders (approved):
Wow that's a lot of output changed 😅.
Nice to have this as a separate file.
Thanks!

PR #231, opened 2023-01-28
226 parsing dependency of goals

Mustafa-Eldeeb:
Parsing goals depending on other goals, using DAG algorithm


tijlleenders (approved):
Great iteration!
Next I'll add a test case to evolve this into evaluating flexibility for task ordering at runtime...

PR #242, opened 2023-02-04
238 planed hierarchy goals

Mustafa-Eldeeb:
We need to review the output before merge we can modify if it differ from expected one.


tijlleenders (approved):
Nice work.
I also like your suggestion that we should let filler goals go last in the task_placer!

PR #245, opened 2023-02-05
244 schedule real tasks first

Mustafa-Eldeeb:
Allow to compute flex for real tasks as filler tasks.
Expected output should be reviewed


tijlleenders:
Resolves #224

tijlleenders (approved):

PR #246, opened 2023-02-05
235 test flex at runtime

### LINKED ISSUE #235: test flex at runtime
For goals being scheduled after other goals , it uses DAG to get write order but not depend on flexibility At runtime.
it should be handled by allowing scheduler based on flexibility At runtime. Without using doNotSort tag.

tijlleenders:
Resolves #235


tijlleenders:
This is the expected logic for reference:
image

tijlleenders:
@Mustafa-Eldeeb You are right the previous picture is wrong.
I've corrected in blue:
image

tijlleenders:
You can merge your own PR since I initiated this.
Nice work!

Mustafa-Eldeeb (approved):

PR #264, opened 2023-02-28
261 reduce weekly budget prorata if start is not on monday

Mustafa-Eldeeb:


tijlleenders:
Discontinued in favor of https://github.com/tijlleenders/ZinZen-scheduler/pull/269

PR #269, opened 2023-03-03
Refactor schedule

tijlleenders:


Tushar-4781 (approved):
This will always be a suspense XD

PR #273, opened 2023-03-23
272 handle budget with no children

Mustafa-Eldeeb:


tijlleenders (approved):
Test case wasn't failing but removed the min-budget suffix so the UI goal titles don't change.

Issue #284, opened 2023-04-01
Filter for days of the week


tijlleenders:
Modify the 'on' filter that now takes Weekends/Weekends/Mondays etc to take an array of max 7 day values.

So if I want to filter on Mondays I would pass 'on' filter ["Mon"]

If I want to filter on weekends I would pass 'on' filter ["Sat","Sun"]

If I want to filter on weekdays I woukd pass 'on' filter ["Mon","Tue","Wed","Thu","Fri"]

First step: make the test cases with input and output jsons.

PR #285, opened 2023-04-03
feat: filter for single or many days of the week

moaz-mokhtar:
Feature to modify the 'on_days' filter that now takes `Weekends`, `Weekends`,  `Mondays`, etc to take an array of max 7 day values.

So if I want to filter on:
- Mondays I would pass 'on' filter ["Mon"]
- Weekends I would pass 'on' filter ["Sat","Sun"]
- Weekdays I would pass 'on' filter ["Mon","Tue","Wed","Thu","Fri"]

**Plan**
- [x] Create `json` test cases
- [x] Initial investigation 
- [x] Develop feature
- [x] Test feature 

---
Resolves issue https://github.com/tijlleenders/ZinZen-scheduler/issues/284


tijlleenders (approved):
Awesome @moaz-mokhtar !
Very readable and structured code. Thanks!

PR #292, opened 2023-04-07
ref: basic refactor into modules

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/291 

---

Basic refactor into modules.


tijlleenders (approved):
Thanks, a lot clearer now!

Issue #289, opened 2023-04-07
Feature request: change default scheduling to not splitting consequent slots


moaz-mokhtar:
When scheduling "Sleep 8h 21-6" the scheduler will schedule your sleep from 21-23 and 0-6, which is obviously not optimal.

We need a way to tell the scheduler not to split certain tasks while scheduling. This could be a do_not_split flag in the beginning - but could evolve to min_scheduled_duration - so probably call it that and set it to the min_duration in the test cases - and let ot default to 1h if not specified.

So a min_scheduled_duration of 8 will tell task_placer that it cannot schedule the task unless it is in a slot of 8 consecutive hours.

First idea for this approach:
- ~~always merge slots into one slot if they are adjacent ; this will for example be necessary after the filter pass which at least splits slots into days~~
- ~~remove time slots that cannot fit min_scheduled_duration before calculating flexibility~~

**Improved approach:**  
We first need to fix the fact that Slots are being split before being able to only schedule them if they fit entirely in one slot.
- Don't split slots at all. Keep the Timeline intact and only remove slots - so that no 'touching' slots can ever appear.
- When creating Tasks with `repeat` value, if beforetime < after time: change start of daily/weekly iteration to take few hours in previous day relative to calendar start. So "Sleep 22-8 daily" should start interating in 24h steps on the day before start of the calendar at 22:00, not 0:00.
- After this, we can add logic in the task_placer to default to not allow splitting - and only allow splitting if optional min_split_duration and/or max_split_duration input is present.

moaz-mokhtar:
@tijlleenders Seeking your support for more info

tijlleenders:
@moaz-mokhtar Updated description. Call me if you need further discussion.

moaz-mokhtar:
I think this is should be the default behavior, and flag for the split. What is your opinion?

tijlleenders:
Agree!

moaz-mokhtar:
Seeking confirmation: 
- System by default will not split consequent tasks which have been split by date.
- This will be for kind of tasks

tijlleenders:
Yeah agree that time slots of a task that are next to each other should be 'glued' into one bigger time slot.
That should probably take care of the problem.

moaz-mokhtar:
This is initial `input.json` and `output.json` which expected the new functionality will proceed and output.

`input.json` :
```json
{
    "startDate": "2023-04-01T00:00:00",
[...]

PR #295, opened 2023-04-12
feat: change default scheduling to not splitting consequent slots 

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/289 
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/312 
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/309 

---

## Plan
- [x] Discussing functionality design
- [x] Initial design
- [x] Initial test cases
- [x] Meeting to change design and adding new feature `Timeline` struct
- [x] Develop new feat `Timline`
- [x] Develop unit tests for `Timeline` methods
- [x] Create new methods for filter operations to be used with `Timeline`
- [x] create unit tests for filter operations
- [x] Create unit tests for Slot subtraction
- [x] Fixing Slot subtraction logic
- [x] Implement `Timeline` inside `TimeSlotIterator` for field `timeline` instead of `Vec<Slot>`
- [x] Investigate results between old `apply_filter` and new `apply_filter`
- [x] Fixed issue caused by `remove_slots` function
- [x] Created more test cases for `remove_slots` function
- [x] Fixed issue in `Iterator` for `TimeSlotIterator`
- [x] Fixed issue in new filter function `filter_not_on` which now supports splitting slots into 1 hour slot and using `retain` function instead of `remove_slots`.
- [x] Develop a function to merge consequent slots (which splitting happen after `filter_not_on` or other)
- [x] Develop test cases for the function merge consequent slots
- [x] Discuss how to avoid `split` into days or hours
- [x] Developed `SlotIterator` with basic test cases. Developed with 2 functions: walk through a slot with custom interval, get count of a slot based on customer interval. 
- [x] Developed `TimelineIterator` with basic test cases which have 2 functions like `SlotIterator`.
- [x] Priority to fix `company use case` which budgeting instead  of forcing Monday-Monday, it should calculate week from the start of the task date.
- [x] Refactor function for weekly repetition to avoid make week from Mon-Mon
- [x] Fix `output.json` for many tests failed after changing weekly repetition mechanism.
- [x] Create test cases for issues https://github.com/tijlleenders/ZinZen-scheduler/issues/309 and https://github.com/tijlleenders/ZinZen-scheduler/issues/312 
- [x] Debugging test case `i309` related to issue https://github.com/tijlleenders/ZinZen-scheduler/issues/309 and found no issues 
- [x] Debugging test case `i312` related to issue https://github.com/tijlleenders/ZinZen-scheduler/issues/312 and found no issues
- [x] Refactor `SlotIterator` and `TimelineIterator` to support walk through slots and custom interval duration and get out list of slots to do required operations like filtering, splitting, etc.
- [x] Test cases added to test walk through slots and extract list of slots based on an interval duration and do some operations lilke filtering.
- [x] Refactor filtering functions to use `SlotIterator` and `TimlineLine` iterators
- [x] Test cases for filtering functions after implementing `SlotIterator` and `TimlineLine` iterators 
- [x] Apply new filtering function instead of old one. Test functionalities.
- [x] Fix function `filter_timing` to solve edge cases
- [ ] Refactor `filter_not_on` to follow the same concept of new implementation of `filter_timing`. Check commit 67578be75ab1f7fba10fd38ab7c5ac66291357ee
- [ ] Refactor `filter_on_days`  to follow the same concept of new implementation of `filter_timing`. Check commit 67578be75ab1f7fba10fd38ab7c5ac66291357ee
- [ ] Refactor splitting functions in the system to use `SlotIterator` and `TimlineLine` iterators instead of old style 
- [x] Fixed `remove_slots` and edge cases passed
- [x] Fixed logic for `filter_not_on` and edge cases passed


tijlleenders:
@moaz-mokhtar I realize the comments in commit aren't entirely accurate. The time slots for a daily habit that are consecutive/touching at midnight will be separated into different Tasks.

The solution is to adapt the task_generator for daily habits. Whenever there is a filter for after that is greater than the before time the entire time slot (so starting before midnight and ending after midnight) should be the basis for the tasks.

So instead of each task having two time slots of 0-8 and 22-0 (on same day) - each task will have ONE time slot of 22-8 spanning TWO days.

tijlleenders:
A daily habit is more about the interval than the calendar day.

moaz-mokhtar:
After latest commit: https://github.com/tijlleenders/ZinZen-scheduler/pull/295/commits/7dbe277fe6c5bd089d9c3a58aa4a7bd35ece55af, after new function `merge_consequent_slots`, found most of the tests are failed; 8 passed only and 71 failed.

tijlleenders:
About https://github.com/tijlleenders/ZinZen-scheduler/commit/7d62372676512992abb1188284fa1aced31e3f6b

The splittable indicator from the frontend has no business in the task_generator/timeslots_iterator - it is only relevant in the task_placer.

tijlleenders (changes requested):
Overall delivers on the requested functionality. A few changes required though.

src/models/timeline.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// - ? Should remove similar slots, or remove from each slot in timeline!!
+    /// - ? Remove from each slot in the timeline means remove one slot from many
+    fn remove_slots(&mut self, slots_to_remove: Vec<Slot>) -> Option<Vec<Slot>>;
+    /// Get a slot of timeline based on index
+    /// - If index is out of range, return None
+    fn get_slot(&self, index: usize) -> Option<Slot>;
+}
+
+impl TimelineOperations for Timeline {
+    fn remove_slots(&mut self, slots_to_remove: Vec<Slot>) -> Option<Vec<Slot>> {
```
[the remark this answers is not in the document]
moaz-mokhtar: Logically that is right which changes will be for `self`, but as per agreed design, `remove_slots` should return slots after remove unwanted slots.
Now I will remove return value.

src/tests/timeline.rs
```diff
@@ -0,0 +1,147 @@
...
+    if let Some(timeline) = Timeline::initialize(sample_slot.start, sample_slot.end) {
+        assert_eq!(expected_timeline, timeline);
+    } else {
+        assert!(false);
+    }
+}
+
+#[test]
+fn test_remove_from() {
+    // sample_slot: [2022-10-01 05:00:00 --- 2022-10-01 20:00:00]
```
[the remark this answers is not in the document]
moaz-mokhtar: It is important to know the return values specially this function return a static data.
But anyways this is temporary, because I will refactor these functions into flexible functions which accepting params

src/tests/timeline.rs
```diff
@@ -0,0 +1,147 @@
...
+    let sample_slot = get_slot_1();
+
+    if let Some(mut timeline) = Timeline::initialize(sample_slot.start, sample_slot.end) {
+        // slot_to_remove: [2022-10-01 05:00:00 --- 2022-10-01 10:00:00]
+        let (slot_to_remove, _) = get_2_slots();
```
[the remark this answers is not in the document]
moaz-mokhtar: No, I meant to return 2 slots with static data.

tijlleenders:
I like your discipline of adding unit tests before developing the code!

TDD is great for getting results - and (at least for me)  developer experience.

moaz-mokhtar:
> I like your discipline of adding unit tests before developing the code!
> 
> TDD is great for getting results - and (at least for me) developer experience.

Thanks, and totally agree.

tijlleenders:
>- [ ] Resolve issue in function `adjust_min_budget_tasks`, check debug feedback on https://github.com/tijlleenders/ZinZen-scheduler/issues/300#issuecomment-1528727445 

I removed this from PR description as it can be solved in a separate PR

src/models/task/impls.rs
```diff
@@ -82,6 +88,47 @@ impl Ord for Task {
...
+    /// - status: status of the task
+    /// - timeframe: Start and deadline of a task
+    pub fn new(
+        task_id: usize,
+        title: &str,
+        duration: usize,
+        goal: &Goal,
+        timeline: &Timeline,
+        status: &TaskStatus,
+        timeframe: Option<Slot>,
```
tijlleenders, line 107: @moaz-mokhtar I don't see why this is required when you have the `timeline`.  
If it's temporary for refactoring please indicate this in code.

moaz-mokhtar: I did create this function for below:
- the challenge to avoid splitting slots outside `Timeline`, which I found 8 times slots splitting outside Timeline; 2 of them are repeated. So I'm extract functions outside calls inside iterator or other functionalities which splitting slots.
- as agreed before that create new task should be from a function not inside `Budget` or other staff; so I found ~ 4 calls for creating new `Task` and I just created a function to be common. It can be better

moaz-mokhtar:
Regarding make splitting `Slot`s only inside `Timeline` we have challenges, I have notes as below:
- 8 times slots splitting outside `Timeline`,
- 2 of them are the same functions called 2 times from functions `TaskBudget::generate_budget_min_and_max_tasks(...)` and from `Goal::generate_tasks(...)`
- `Timeline` will not handle it by only `remove_slots(Vec:<Slots>)`. Imagine there is a timing filter, how to ask `Timeline.remove_slots`? How to ask `Iterator` for `TimeSlotIterator` not split the slots.
- So what I'm doing is divide and conquer, until reach ability to avoid any splitting or playing with `Slot`s from outside `Timeline`

src/models/task/impls.rs
```diff
@@ -82,6 +88,47 @@ impl Ord for Task {
...
 impl Task {
+    /// Create new task
+    /// ## Parmeters:
+    /// - task_id: id of the task
+    /// - title: title of the task
+    /// - duration: duration of the task
+    /// - goal: goal of the task
+    /// - timeline: timeline of the task
+    /// - status: status of the task
+    /// - timeframe: Start and deadline of a task
```
tijlleenders, line 99: This way of documenting is used in the [Rust docs](https://doc.rust-lang.org/rust-by-example/meta/doc.html) - so I understand why we would try this. However - this documentation style is not used in the standard library.  

I think I err on the side of the standard library. **A function doc comment should be restricted to a simple phrase in the general description, and possibly a thing or two that is important to pay attention to.**  
Arguments:  
- Clean code guidelines state the number of params should be kept small - so a short description is easy.  
There is no need for a long list of params that need documentation.  
- Documenting the parameters is currently not auto-checked and thus leads to out-of-date docs.  
- If the params are structs, the docs are in the struct. Providing a summary would lead to duplication.

src/models/task/impls.rs
tijlleenders: Ok I understand now it's about standardizing before doing further refactoring. Perfect!

src/models/task/impls.rs
moaz-mokhtar: I fixed that  by encapsulating the many parameters into single struct `NewTask`. Check and give me a feedback

tijlleenders: Well, the doc comments are gone now so yeah!  
The `NewTask` was not necessary in my opinion - it doesn't really add anything except indirection - but I guess it's temporary anyway so good.

tijlleenders:
@moaz-mokhtar So since all the tests are passing right now and the company use case is important to fix for the demo video's I've added it to this PR.  

**Problem:**  
Right now we use 'calendar weeks' for budgets - so always Monday-Monday. This is why a 'work 40h/week' budget is failing when requesting a schedule starting on Friday; one workday is not enough to fill the weekly budget.

**Proposed solution:**  
We always calculate the weekly budget as of the starting datetime + one week.  
So if a weekly budget of 'work 40h/week' is requested on a Friday, it will only generate **one budget**:
- Friday-Friday without any impossible
instead of  
- one Friday-Monday budget that causes impossible Tasks
- a second Monday-Friday budget that schedules successfully

moaz-mokhtar:
> @moaz-mokhtar So since all the tests are passing right now and the company use case is important to fix for the demo video's I've added it to this PR.
> 
> **Problem:** Right now we use 'calendar weeks' for budgets - so always Monday-Monday. This is why a 'work 40h/week' budget is failing when requesting a schedule starting on Friday; one workday is not enough to fill the weekly budget.
> 
> **Proposed solution:** We always calculate the weekly budget as of the starting datetime + one week. So if a weekly budget of 'work 40h/week' is requested on a Friday, it will only generate **one budget**:
> 
>     * Friday-Friday without any impossible
>       instead of
> 
>     * one Friday-Monday budget that causes impossible Tasks
> 
>     * a second Monday-Friday budget that schedules successfully


The `Mon to Mon` weekly repetition now fixed to be for a week. I did refactor this functionality into a separate function with test cases for many days of the week.
But the current `output.json` in the test folder `demo-3-with-filler-and-budget`, shows a result that may happen when resolving  issue https://github.com/tijlleenders/ZinZen-scheduler/issues/300.
Also I did create a correct `output.json` based on current impl in folder named `demo-3-ver2`. The difference is that this test shows scheduling based on when `min_duration` for week achieved, will not add anymore.

moaz-mokhtar:
I found below tests affected by new weekly repetition mechanism, so I'm going to fix the `output.json` for them.

```bash
failures:
    demo_2_with_filler_without_budget
    demo_3_with_filler_and_budget
    different_repetition_goals
    flex_duration_1
    flex_duration_2
    flex_repeat_1
    flex_repeat_2
    flex_repeat_3
    issue_276_weekdays_filter_on_budget
    show_sleep
    weekly_1
    weekly_2
    x_per_week_2
    x_per_week_3

test result: FAILED. 64 passed; 14 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.86s

```

tijlleenders:
@moaz-mokhtar Thanks for the commits!

1. Yeah so we can ignore `demo_3_with_filler_and_budget` failing till #300 gets fixed. When that happens we can remove `demo-3-ver-2` since that is the temporary expectation without max-budgets implemented.  
2. `demo_2_with_filler_without_budget` input is identical to `demo_3_with_filler_and_budget` so can be removed.
3. For the other tests indeed, please correct the expectation. I expect the diff to be that there are now no more impossible tasks in the outputs - or simply different timings/naming due to the different starting/ending times for the week.
4. Observation: logging an issue and using that issue number in the test name whenever we need to add a scheduler tests is a good practice. This provides the context that is otherwise hard to find just looking at the input.

src/models/timeline/iterator.rs
```diff
@@ -0,0 +1,193 @@
...
+                interval: interval_duration,
+            }
+        } else {
+            panic!("Timeline slots are empty")
+        }
+    }
+
+    /// Get count between start and end for all slots in
+    /// timeline based on Duration interval
+    pub fn interval_count(&self) -> usize {
```
tijlleenders, line 57: Why is this functionality needed?  
I don't see it being called anywhere?

src/models/slots_iterator/utils.rs
```diff
@@ -325,28 +325,13 @@ pub fn get_start_of_repeat_step(
...
-    panic!("shouldn't reach")
+    return result
+        .checked_add_days(Days::new(7))
+        .unwrap()
+        .with_hour(0)
+        .unwrap()
+        .with_minute(0)
+        .unwrap()
+        .with_second(0)
+        .unwrap();
```
tijlleenders: This is not consistent with the commit title. It is actually the fix for the added (failing) tests.

src/models/timeline/iterator.rs
moaz-mokhtar: Actually I found many functions which not called yet, so I'm going to remove them gradually when no need for them.

moaz-mokhtar:
I did remove `demo_3_with_filler_and_budget` and `demo_3_ver_2`, then just kept demo-2 and renamed to `demo-2-with-filler-with-budget`

src/services/timeline.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// Filtering timeline based on not_on field in TimeFilter
+    pub fn filter_not_on(&mut self, _slots_to_filter: &[Slot]) {
+        todo!("filter_not_on not implemented");
+    }
+}
+
+/// Validate that a given value is valid time number which must be between 0 and 24
+fn _validate_time(time: Option<usize>, time_name: &str) {
```
tijlleenders, line 29: Good to add this check!  
Isn't a more appropriate place to do this when the input gets deserialized?

src/services/timeline.rs
```diff
@@ -0,0 +1,78 @@
...
+    timeline::Timeline,
+};
+
+impl Timeline {
+    pub fn apply_filter(&mut self, _filter: &Option<TimeFilter>) -> Timeline {
+        todo!("apply_filter not implemented");
+    }
+
+    /// Filtering timeline based on before_time and after_time fields in TimeFilter
+    pub fn filter_timing(&mut self, _before_time: Option<usize>, _after_time: Option<usize>) {
```
tijlleenders, line 13: Since we have `apply_filter' exposed towards the outside, the `filter_*` methods can be private?

moaz-mokhtar: Yes, that is good design, thanks

src/services/timeline.rs
moaz-mokhtar: That will be better, to refuse in appropriete/wrong input from users. But I think this will be developed when adding error handling features like Issue https://github.com/tijlleenders/ZinZen-scheduler/issues/299

tijlleenders:
Getting there!

tijlleenders:
Great @moaz-mokhtar !

Actually I think the old output for `every_2_hours_overlap` and `sleep_1` should be maintained - as the daily iteration should start on the day **_before_** at the `after` time, as stated in https://github.com/tijlleenders/ZinZen-scheduler/issues/289#issuecomment-1531485611

moaz-mokhtar:
> Great @moaz-mokhtar !
> 
> Actually I think the old output for `every_2_hours_overlap` and `sleep_1` should be maintained - as the daily iteration should start on the day **_before_** at the `after` time, as stated in [#289 (comment)](https://github.com/tijlleenders/ZinZen-scheduler/issues/289#issuecomment-1531485611)

Thanks,

For the comment https://github.com/tijlleenders/ZinZen-scheduler/issues/289#issuecomment-1531485611, that is implemented by new filter function `filter_timing` , check test `test_beforetime_is_before_aftertime` https://github.com/tijlleenders/ZinZen-scheduler/blob/0f5c3da3a8cf58e622de80b5a6d8d0045ec24723/src/services/filter/filter_timing.rs#L211

tijlleenders:
> For the comment [#289 (comment)](https://github.com/tijlleenders/ZinZen-scheduler/issues/289#issuecomment-1531485611), that is implemented by new filter function `filter_timing` , check test `test_beforetime_is_before_aftertime`

@moaz-mokhtar That expectation is wrong too. It should have a first slot on 20230501 from 0:00-5:00.
Please use before_time and after_time instead of start_time and end_time - they have different meanings.

tijlleenders:
@moaz-mokhtar After discussion the way you're doing timenline filtering is easier to reason about.  
So for the filter implementation let's iterate over the timeline slots and keep what we want - instead of figuring out what we don't want and then removing it.

moaz-mokhtar:
> @moaz-mokhtar After discussion the way you're doing timenline filtering is easier to reason about. So for the filter implementation let's iterate over the timeline slots and keep what we want - instead of figuring out what we don't want and then removing it.

For the `filter_timing`, I found this solution is more convenient at this time, but for the other 2 filter functions `filter_not_on` and `filter_on_days`, it will be more complex and time consuming at the moment to use `TimelineIterator` specially with `filter_not_on` which it simpler to use Slot substraction with `remove_filters` instead of `TimelineIterator`.
For the the function `filter_on_days`, I didn't analysis it, but I think it will not be complex to use `TimelineIterator` Check commit 5d405e2157eb63ab6aa7de4824df50d96d3f828b

src/models/budget/impls.rs
```diff
@@ -193,31 +170,65 @@ impl TaskBudgets {
...
 
-            for time_slots in time_slots_iterator {
+            for timeline in time_slots_iterator {
                 let task_id = *counter;
                 *counter += 1;
-                if !time_slots.is_empty() {
-                    let task = Task {
-                        id: task_id,
-                        goal_id: goal.id.clone(),
+                if !timeline.slots.is_empty() {
```
tijlleenders, line 176:
```suggestion
                if timeline.slots.len() > 0 {
```
Clean code standard to not use (double) negatives in 'if' statements.
It is clearer when you say: only use postitives in 'if' statements. Any negative can be converted to a posititve.

src/models/goal/impls.rs
```diff
@@ -0,0 +1,154 @@
...
+            // Todo! add self.before_time filter
+        );
+        dbg!(&time_slots_iterator);
+
+        for timeline in time_slots_iterator {
+            dbg!(&timeline);
+            let task_id = *counter;
+            *counter += 1;
+            // TODO 2023-05-06  | apply Task::new(...)
+            if !timeline.slots.is_empty() && self.min_duration.is_some() {
```
tijlleenders, line 120:
```suggestion
            if timeline.slots.len() > 0 && self.min_duration.is_some() {
```

tests/jsons/flex-duration-2/output.json
```diff
@@ -858,7 +858,16 @@
...
+      "outputs": [
+        {
+          "taskid": 98,
+          "goalid": "3",
+          "title": "sleep",
+          "duration": 2,
+          "start": "2022-10-01T00:00:00",
+          "deadline": "2022-10-15T00:00:00"
+        }
+      ]
```
tijlleenders: @moaz-mokhtar It would be nice if this test case is fully possible (no impossible).  

A way to do that is by disregarding the calendar start/end in the goal's timeline and task_placer.  
So the sleep timeline would start on either:
- the goal.start if present - adjusted to the after time
- one day before calendar start - adjusted to the after time - if no goal.start is present

The calendar start/end would be taken into account in the output formatter - which would then 'chop off' the first day.

tijlleenders: We'll merge this PR and create a separate issue for this.

tijlleenders (approved):
Wow!
This is more like a dev branch than a PR. Let's try to make small separate issues and merge frequently instead :)

src/models/budget/impls.rs
moaz-mokhtar: I think I will return it back as below warning by `cargo clippy`:
```rust
❯ cargo clippy
warning: length comparison to zero
   --> src/models/budget/impls.rs:176:20
    |
176 |                 if timeline.slots.len() > 0 {
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!timeline.slots.is_empty()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero
    = note: `#[warn(clippy::len_zero)]` on by default
```

src/models/goal/impls.rs
moaz-mokhtar: Also Warning happen here by `cargo clippy` :

```bash
warning: length comparison to zero
   --> src/models/goal/impls.rs:120:16
    |
120 |             if timeline.slots.len() > 0 && self.min_duration.is_some() {
    |                ^^^^^^^^^^^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!timeline.slots.is_empty()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero

warning: `zinzen` (lib) generated 2 warnings (run `cargo clippy --fix --lib -p zinzen` to apply 2 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
```

tijlleenders: Ha yeah I don't _always_ agree with Clippy.

Do as you like:  
- use `!is_empty`
- turn this Clippy warning off.

moaz-mokhtar:
This Pull Request closed due to big of changes outside the context of the main issue. For the main issue, we fixed not splitting consequent slots from the filtering functions part.

PR #296, opened 2023-04-14
Update issue templates

tijlleenders:
Added issue templates.
Let's use this for a while and iterate!


moaz-mokhtar (approved):
Good idea. Thanks

PR #302, opened 2023-05-02
Update/docs update

tijlleenders:
Hi @moaz-mokhtar ,
As discussed some initial documentation - separate from #295 

Let's have a call soon to discuss any questions/concerns/todos you might have!

### DIFF
--- a/src/bin/test_for_docs.rs
+++ b/src/bin/test_for_docs.rs
@@ -1,0 +1,24 @@
+extern crate scheduler;
+use scheduler::models::input::Input;
+
+fn main() {
+    let json_input: serde_json::Value = serde_json::json!({
+      "startDate": "2022-01-01T00:00:00",
+      "endDate": "2022-01-09T00:00:00",
+      "goals": {
+        "uuid1": {
+          "id": "uuid1",
+          "title": "sleep",
+          "min_duration": 8,
+          "repeat": "daily",
+          "filters": {
+            "after_time": 22,
+            "before_time": 8
+          }
+        }
+      }
+    });
+    let input: Input = serde_json::from_value(json_input).unwrap();
+    let output = scheduler::run_scheduler(input);
+    dbg!(output);
+}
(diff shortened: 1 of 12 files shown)


src/bin/test_for_docs.rs
moaz-mokhtar, line 1: What benefit of this function?

src/models/task.rs
```diff
@@ -5,31 +5,40 @@ use chrono::NaiveDateTime;
...
     pub status: TaskStatus,
+    /// Used for finding next Task/Increment to be scheduled in combination with Task/Increment Status, after_goals and Tags.
     pub flexibility: usize,
+    /// Final start time for Task/Increment on Calendar - should be removed in favor of Timeline + SlotStatus combination.
     pub start: Option<NaiveDateTime>,
+    /// Final end time for Task/Increment on Calendar - should be removed in favor of Timeline + SlotStatus combination.
     pub deadline: Option<NaiveDateTime>,
+    /// The places on Calendar that could potentially be used given the Goal constraints - and what other scheduled Tasks/Increments already have consumed.
     pub slots: Vec<Slot>,
+    /// Used for finding next Task/Increment to be scheduled in combination with Task/Increment flexibility, after_goals, and Status.
```
moaz-mokhtar, line 33: Is that the definition of Tags, which it seems the same definition of `after_goals`!

moaz-mokhtar (changes requested):
Thanks, this will minimize efforts to understand some complex parts of the system.

src/models/task.rs
tijlleenders: Fixed!

src/bin/test_for_docs.rs
tijlleenders: Just copied it from an example doc.
According to https://users.rust-lang.org/t/usage-of-extern-crate/73619 :
>extern crate was necessary in the 2015 edition to actually use a dependency you declared in Cargo.toml. Since the 2018 edition crates will implicitly be loaded at the first reference. Using extern crate is thus not necessary anymore in (almost) all cases and is advised against.

So I removed it.

moaz-mokhtar (approved):
Thanks

PR #304, opened 2023-05-03
remove dynamically generated test file

tijlleenders:
`/tests/rust_tests.rs` was accidentally pushed to GitHub.  
It changes on every build - so I included it in .gitignore - but it also has to be removed from the repo.


tijlleenders:
Actually, after reading https://internals.rust-lang.org/t/pre-rfc-dynamic-tests/6790 and https://internals.rust-lang.org/t/dynamic-tests-revisited/18095  

Let's keep generating the tests - but...:
- make the order deterministic so it only changes minimally when new test cases are added
- since clippy is complaining and docs.rs build is failing: move it out of the build script into a separate script

PR #305, opened 2023-05-03
rm unused buil-dependency

tijlleenders:


moaz-mokhtar:
This removed dependency related to build scripts `build.rs`, but I see the same behavior. So why do you think this is important?

tijlleenders:
> This removed dependency related to build scripts `build.rs`, but I see the same behavior. So why do you think this is important?

I don't think this crate is being used. That's why I removed it.  
It is not related to `build.rs` - just cleaning up.

Cargo.toml
```diff
@@ -47,7 +47,6 @@ soft = "0.1.1"
 
 # used to generate tests
```
Mustafa-Eldeeb, line 48:
```suggestion
```

Cargo.toml
```diff
@@ -47,7 +47,6 @@ soft = "0.1.1"
...
 [build-dependencies]
```
Mustafa-Eldeeb, line 49:
```suggestion
```

Mustafa-Eldeeb (commented):
You can remove CC from build deps. Also I suggest to clean the whole section

tijlleenders (commented):
Ok please merge.

Mustafa-Eldeeb (approved):

Issue #306, opened 2023-05-03
Don't generate dynamic tests from build.rs


tijlleenders:
**Describe the bug**
Docs.rs build is failing to build.   
It is also annoying that the dynamic tests are in a different order after every build - it causes a lot of git diffs.  

**To Reproduce**
1. `cargo publish --dry-run`
4. See error :  
```
error: failed to verify package tarball  

Caused by:  
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.  
  Changed: /home/tijl/ZinZen-scheduler/target/package/zinzen-0.1.1/tests/rust_tests.rs  
```

**Expected behavior**
I can cargo publish without the build failing - and tests don't change if input doesn't change.  

**Solution**  
Let's keep generating the tests - but...:
- [x] make the order deterministic so it only changes minimally when new test cases are added.  
- [ ] Move dynamic tests generation out of the build script to `/src/bin/generate_rust_tests.rs`.   
We could have a 'shadow tests' directory with inputs and outputs so that any changes can easily be viewed in change control?
- [ ] In build.rs invoke `/src/bin/generate_rust_tests.rs` without writing to disk (using a no-write flag) - just compare with the one on disk and panic if there is a difference! This will catch it locally as well as during CI - but also allow publishing to docs.rs - as build.rs never writes outside of the OUTPUT_DIR.

**Alternate solution**  
- Put all the tests into one dynamic test. Not ideal for several reasons explained in links below.  
- Do some trickery with other crates mentioned in links below. Too complex compared to the 'good enough' solution proposed under 'Solution' header.  

**Context**
https://internals.rust-lang.org/t/pre-rfc-dynamic-tests/6790 and https://internals.rust-lang.org/t/dynamic-tests-revisited/18095

moaz-mokhtar:
Check if issue minimized or resolved in PR: https://github.com/tijlleenders/ZinZen-scheduler/pull/307; I did force sort test directories so it will not be changed else if directories have changed.

I suggest below:
- no need to force this test at the beginning, so tests will be generated with a script manually
- make this test by `npm` only and let test in Rust as it going with PR https://github.com/tijlleenders/ZinZen-scheduler/pull/295

tijlleenders:
> make this test by npm only  
[...]

PR #307, opened 2023-05-04
test: unsorted dynamic tests from buildrs

moaz-mokhtar:
Only resolves part of #306 

---


tijlleenders (approved):
Thanks, this fixes the most pressing part of #306  
I've edited the description to show there's more work to be done.  

The rest can wait - it's low prio compared to the actual scheduler.

PR #322, opened 2023-06-10
fix: fixed Slot Sub rule when rhs large than slot

moaz-mokhtar:
This is a sub-PR for PR https://github.com/tijlleenders/ZinZen-scheduler/pull/320

--- 

Fixed Slot Sub trait impl when rhs > slot. In old implementation it is swapped and subtract larger slot from smallest slot, but now after fix it just return empty list.


tijlleenders (approved):
Thanks, this is what any unsuspecting user of Sub would expect :)

PR #325, opened 2023-06-17
fix: test case flex-repeat-2

moaz-mokhtar:
This is a sub-PR for PR https://github.com/tijlleenders/ZinZen-scheduler/pull/320

--- 

Fix test case `flex-repeat-2` and related functions.

--- 
Update:
Agreed to remove all flex repeat tests which no priority at the moment for it.


moaz-mokhtar:
- I want more info, how flex calculated for `FlexWeekly` goals?
- Need confirmation what the rule of `FlexWeekly` because implementation is different from theory:
  - theory: "If given a goal with 3-5x/week, create 3 goals and 2 extra optional goals"
  - implementation: If given a goal with 3-5/week, create 3 goals and 1 extra optional goals (5-1). check below code snippet:
Note: function `add_optional_flex_number_and_duration_habits_goals` renamed to `generate_flex_weekly_goals`
  ```rust
  fn add_optional_flex_number_and_duration_habits_goals(goals: &mut GoalsMap) {
      // ...
      for (goal_id, goal) in goals.iter_mut() {
          if let Some(Repetition::FlexWeekly(min, max)) = goal.repeat {
              // ,..
              for number in min..max - 1 { // << here will decrease 1 optional goal
                  // ...
              }
              // ...
          }
      }
      // ...
  }
  ```

tijlleenders:
@moaz-mokhtar 
Looks like the implementation is a typical off by one error. I'm not sure if using normal + optional goals is the right strategy either... but something more like a budget will take us into some deep woods.  

Anyway - the UI currently doesn't support flex-repeat so I think we should not put too much effort in this. If the off-by-one fixes it the  fine ; otherwise we can park this and focus on thr next failing test.

tijlleenders:
@moaz-mokhtar 
image
This calculation sheet is not accurate.
1. Let's start by talking about Steps instead of Goals. A daily Goal results in one Step per day. Flexibility is never calculated on a Goal - only on a Step. So the Goal 'Family time' can't have a flexibility. Only the Steps thats result from the Goal have a flexibility. In this case the flexibility for each Step is 1 - so the Steps from this Goal probably will get scheduled first.
2. The input needs to be corrected. "weekdays" for Goal "exercise" is not a valid repeat value. Repetition is a simple concept that means "withing the Goal timeline, create a new Step every x duration from the startdatetime". So what needs to be done is modify the `enum Repetition` to only allow for `EveryXhours(usize)`, and the convenience enum `EveryXdays(usize)` to make things more readable. The rest can be removed - except for the flex-repeat `FlexDaily(usize, usize)` and `FlexWeekly(usize, usize)`.
3. In my previous comment I talked about `Goal` when I should have said `Step`. The flex-repeat Goal 'side project' with repeat '1-3/week' is transformed into three Steps _for every week_ in the Goal timeline - so 1 regular Step and 2 optional Steps for each week.
4. Just FYI, to keep things simple, a Goal with a repetition can't have children - for now. As far as I know we respect that it the current test cases.

moaz-mokhtar:
> Anyway - the UI currently doesn't support flex-repeat so I think we should not put too much effort in this. If the off-by-one fixes it the fine ; otherwise we can park this and focus on thr next failing test.

FYI:
- Need a base to resolve this issue, current  `output.json` is not a valid one!
- So we should put more effort to get accurate `output.json` by excel sheet scheduling simulator and other ways.
- the other test case `flex-repeat-3` seems to have a common with this test case for this PR.
- In case we park this test case and expect the other one `flex-repeat-3`, so I can propose to put all `json` tests which doesn't have a priority like this  into a separate directory to focus efforts on priorities.

tijlleenders:
@moaz-mokhtar 
Just delete all the flex repeat tests.  
We have version control.

moaz-mokhtar:
@tijlleenders 
Should I consider tasks which contains `"repeat": "1/week"`, `"repeat": "4/week"`, or similar?

tijlleenders:
@moaz-mokhtar 
> Should I consider tasks which contains `"repeat": "1/week"`, `"repeat": "4/week"`, or similar?

These are all inputs that currently cannot be configured by the UI so they can be removed. Please make an issue that mentions adding these tests back when we're ready for it. Mention the specific commits so it's easy to revert them ie https://github.com/tijlleenders/ZinZen-scheduler/commit/0e3bb0ed8f94281b3f0afe87a53603fe3a2ce601

moaz-mokhtar:
Issue created ref: https://github.com/tijlleenders/ZinZen-scheduler/issues/326
 
For tasks with `"repeat": "3/day"`, are those included in the remove process?

tijlleenders:
@moaz-mokhtar 
> For tasks with `"repeat": "3/day"` and `"repeat": "every 2 days"`, are those included in the remove process?
"3/day" remove  
"every 2 day" keep

Issue #333, opened 2023-06-23
fix failed test case `budget-with-no-children`


moaz-mokhtar:
**Describe the bug**
Fix test case `budget-with-no-children`

moaz-mokhtar:
This case to be closed which resolved by PR https://github.com/tijlleenders/ZinZen-scheduler/pull/335

PR #335, opened 2023-06-23
fix: case budget-with-no-children

moaz-mokhtar:
Resolved PR https://github.com/tijlleenders/ZinZen-scheduler/issues/333 

---

Issue: fix failed test case `budget-with-no-children`

Activities
- [x] Rename goal's id into better naming to make it easy to debug
- [x] ...


tijlleenders:
@moaz-mokhtar 
I've added an expectation - you'll have to 'escape' coalescing individual steps from the same goal that are scheduled next to each other into a single Task in the output - otherwise you can't track the individual Steps by title.  
For example, the following should be a single Task in the output since they should all have the same goalid "work" - but I kept them separate for tracking the steps:  
```
{
          "taskid": 104,
          "goalid": "19-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T21:00:00",
          "deadline": "2018-01-06T22:00:00"
        },
        {
          "taskid": 105,
          "goalid": "20-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T22:00:00",
          "deadline": "2018-01-06T23:00:00"
        },
        {
          "taskid": 106,
          "goalid": "21-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T23:00:00",
          "deadline": "2018-01-07T00:00:00"
        }
```

Maybe you should not number the goals - as they are all from the same goal... instead number the titles so you can follow the steps that belong to (are created from) a single goal?

moaz-mokhtar:
Naming is just to make it easy to figure out. No problem to change naming convention.

But for your info, `goal_id` is not gonna be changed. You can separate between `Output` by `task_id`, so it is expected to found many `Output` tasks with same `goal_id` but must be different `task_id`

```
{
          "taskid": 104,
          "goalid": "1-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T21:00:00",
          "deadline": "2018-01-06T22:00:00"
        },
        {
          "taskid": 105,
          "goalid": "1-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T22:00:00",
          "deadline": "2018-01-06T23:00:00"
        },
        {
          "taskid": 106,
          "goalid": "1-work",
          "title": "Work",
          "duration": 1,
          "start": "2018-01-06T23:00:00",
          "deadline": "2018-01-07T00:00:00"
        }
        
        ...
}
```

tijlleenders:
@moaz-mokhtar 
Ok so there is only one goal for work from which many steps are created, right. In the end the stepid disappears and a taskid is generated sequentially?  

...but if there is only one goal for work - why call it '1-work' and not just 'work'?

moaz-mokhtar:
> @moaz-mokhtar Ok so there is only one goal for work from which many steps are created, right. In the end the stepid disappears and a taskid is generated sequentially?
> 
> ...but if there is only one goal for work - why call it '1-work' and not just 'work'?

This is just to make it easy to followup with generated tasks/steps. So in case `goal_id: 1`, so I need to lookup what is `goal_id: 1`, but if it is `goal_id: goal` or `goal_id: 1-goal` that will be better specially when you have alot alot of `Output` tasks.
Also number in `goal_id`  like `1-work`, just to note the order in the `input.json`. 

I hope this way make it easy to figure out cases similar to this

src/services/budgeting.rs
```diff
@@ -34,67 +34,70 @@ impl StepBudgets {
             goals.get_mut(&goal_id).unwrap().tags.push(Tag::Budget);
         }
         //For each budget add all descendants
-        for budget in &self.budget_id_to_budget {
-            let mut parents_to_go: Vec<String> = vec![budget.0.clone()]; //start with the goal that initiates the budget
-            self.goal_id_to_budget_ids
-                .insert(budget.0.clone(), vec![budget.0.clone()]); //add itself for when budget filler min-max need to be checked with budget
+        for (goal_id, _) in &self.budget_map {
```
tijlleenders, line 37: There is no goal_id in budget_map, right?

tijlleenders (approved):

src/services/budgeting.rs
moaz-mokhtar: Actually there is, that why I renamed it to avoid conflicts. Check this: https://github.com/tijlleenders/ZinZen-scheduler/blob/a003845f8895c8d156b5b084e6bf22d9f57fc981/src/models/budget/impls.rs#L74

tijlleenders:
@moaz-mokhtar  
When I run `./test rust_tests` in scripts dir it doesn't run any tests:  
```

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out; finished in 0.00s
```

src/services/budgeting.rs
tijlleenders: Ok I'll revert what I thought was a typo then...

tijlleenders (approved):
@moaz-mokhtar  
I see some issue with the budget maps - but the test case is passing so let me find another one that will fail :)

moaz-mokhtar:
> @moaz-mokhtar When I run `./test rust_tests` in scripts dir it doesn't run any tests:
> 
> ```
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s
> 
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
> 
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out; finished in 0.00s
> ```

Good finding.
Actually I did used test and test for a case

moaz-mokhtar:
> @moaz-mokhtar When I run `./test rust_tests` in scripts dir it doesn't run any tests:
> 
> ```
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s
> 
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
> 
> 
> running 0 tests
> 
> test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out; finished in 0.00s
> ```

If you tried `cargo test rust_tests` you will find the same result. So I'm going to make a named module for tests inside `rust_tests.rs` file, so we can run them seprately

moaz-mokhtar:
> > @moaz-mokhtar When I run `./test rust_tests` in scripts dir it doesn't run any tests:
> > ```
> > 
> > running 0 tests
> > 
> > test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s
> > 
> > 
> > running 0 tests
> > 
> > test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
> > 
> > 
> > running 0 tests
> > 
> > test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out; finished in 0.00s
> > ```
> 
> If you tried `cargo test rust_tests` you will find the same result. So I'm going to make a named module for tests inside `rust_tests.rs` file, so we can run them seprately

Resolved in commit https://github.com/tijlleenders/ZinZen-scheduler/pull/342/commits/97bf79425f95989d839b1866a9e040193790ac06, so you can run `./scripts/test e2e` to run only tests in `rust_tests.rs`

Issue #343, opened 2023-06-30
fix sleep-1


tijlleenders:
**Describe the bug**
Testcase sleep-1 is failing.

**Expected behavior**
Testcase sleep-1 passes.
![image](https://github.com/tijlleenders/ZinZen-scheduler/assets/44887755/2993fb10-175e-4864-b309-356c3db0c91e)

moaz-mokhtar:
Case `sleep-` passed as per PR https://github.com/tijlleenders/ZinZen-scheduler/pull/342.
But issue related to the process of finding conflicted slots, which currently it is just comparing a slot block with another slot block. So now we want to change this behavior, so when comparing a slot block with another, it will compare each hour inside it not just the block, so that leads to more accurate slot conflicts finding like pic in the definition

tijlleenders:
@moaz-mokhtar 
Really? 
![image](https://github.com/tijlleenders/ZinZen-scheduler/assets/44887755/95bf9acc-4e36-4446-9d56-67d6d78ec747)

moaz-mokhtar:
Yes, current process just find conflicts between 2 block slots and not find conflicts for each hour inside them. So based on current process the tests for `split-1`, `split-2`, and `split-3` are valid; that is why I did fix `output.json` for all of them which they are passed.
But after implementing new process as per this issue, it will be different concept and expected to affect another tests.

moaz-mokhtar:
This case resolved `split-1` instead of `sleep-1` by mistake for the similarities of the names

Issue #340, opened 2023-06-29
rename all output jsons


tijlleenders:
**Describe the bug**
`output.json` feels like it is the observed output.  
`actual_output.json` is also ambiguous - is it what you actually expect? Maybe it's just me...

**Expected behavior**
Clear which output.json is expected and which one is observed.  
Rename `output.json` to `expected.json`.  
Rename `actual_output` to `observed.json`.

moaz-mokhtar:
Good note, thanks for that

moaz-mokhtar:
Resolved in PR: https://github.com/tijlleenders/ZinZen-scheduler/pull/344. Check commit: https://github.com/tijlleenders/ZinZen-scheduler/pull/344/commits/dc71f2a8d901b58fe5f9d887240078b68d245e46

PR #342, opened 2023-06-30
fix: split test cases 

moaz-mokhtar:
Fix split test cases `split-1`, `split-2`, `split-3`, and `split-4`.

PR #344, opened 2023-06-30
fix: change find conflicts process

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/343
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/349
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/340

--- 

**Change finding conflicted slots process.**
Currently:
-  It is just comparing a slot block with another slot block which not provide accurate scheduling. Also that is appeared clearly in test case `sleep-1`. Check PR: https://github.com/tijlleenders/ZinZen-scheduler/pull/342

Proposal:
-  So now we want to change this behavior, so when comparing a slot block with another, it will compare each hour inside it not just the block, so that leads to more accurate slot conflicts finding. Check image in https://github.com/tijlleenders/ZinZen-scheduler/issues/343#issue-1782227832


tijlleenders (dismissed):
Great, no regressions.
Observed for following test cases changed, but were failing and are still failing:
- #350 
- #351

tijlleenders (approved):

Issue #347, opened 2023-07-01
rename repeat values that are actually filters


tijlleenders:
**Describe the bug**
Repeat values like "Weekdays", "Mondays", "Weekends" are not actually repeat values - but a filter of "on_days" followed by a repeat every 24h (daily).

**Expected behavior**
Refactor all input files that use the 'old' repeat values.

moaz-mokhtar:
Should I apply this in the `input.json` files only, or will be forced in internally  in the `Goal` struct.

```rust
pub struct Goal {
    ...

    pub repeat: Option<Repetition>,

    ...
}

pub enum Repetition {
    DAILY(usize),
    HOURLY,
    Weekly(usize),
    WEEKDAYS,
    WEEKENDS,
    EveryXdays(usize),
    EveryXhours(usize),
    MONDAYS,
    TUESDAYS,
    WEDNESDAYS,
    THURSDAYS,
    FRIDAYS,
    SATURDAYS,
    SUNDAYS,
    FlexDaily(usize, usize),
    FlexWeekly(usize, usize),
}
```

tijlleenders:
Removing it in the struct will prevent it from being used - so good to do.

I would make that into a separate issue as it could impact other code internally.

PR #353, opened 2023-07-03
test: replaced repeat values in input.json files as requested

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/347

---

Replaced repeat values in input.json files like `Weekends`, `Weekdays` and used instead field `filter.on_days`

### DIFF
--- a/tests/jsons/before-7/expected.json
+++ b/tests/jsons/before-7/expected.json
@@ -22,27 +22,18 @@
         {
             "day": "2022-10-02",
             "tasks": [{
-                    "taskid": 2,
-                    "goalid": "1",
-                    "title": "run",
-                    "duration": 1,
-                    "start": "2022-10-02T00:00:00",
-                    "deadline": "2022-10-02T01:00:00"
-                },
-                {
-                    "taskid": 3,
-                    "goalid": "free",
-                    "title": "free",
-                    "duration": 23,
-                    "start": "2022-10-02T01:00:00",
-                    "deadline": "2022-10-03T00:00:00"
-                }
-            ]
+                "taskid": 2,
+                "goalid": "free",
+                "title": "free",
+                "duration": 24,
+                "start": "2022-10-02T00:00:00",
+                "deadline": "2022-10-03T00:00:00"
+            }]
         },
         {
             "day": "2022-10-03",
             "tasks": [{
-                "taskid": 4,
+                "taskid": 3,
                 "goalid": "free",
                 "title": "free",
                 "duration": 24,
@@ -53,7 +44,7 @@
         {
             "day": "2022-10-04",
             "tasks": [{
-                "taskid": 5,
+                "taskid": 4,
                 "goalid": "free",
                 "title": "free",
                 "duration": 24,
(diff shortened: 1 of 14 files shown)


tijlleenders (changes requested):
On second thought, a weekly repetition would fit better whenever 'Weekdays' or 'Weekends' is used, right?

moaz-mokhtar:
> On second thought, a weekly repetition would fit better whenever 'Weekdays' or 'Weekends' is used, right?

I'm not sure about system behavior. 
- If we said `daily`, it will filter `on_days` from all other week days. 
- If we said `weekly`, i'm not sure about behavior

tijlleenders:
It is simple.  
If I say 'Run 5km weekends' do you expect to run 5km or 10km?  
A daily repeat on a timeline filtered for Sat + Sun would result in 2 steps (2x5km) - a weekly repeat would result in 1 step (1x 5km with a timeline of Sat + Sun).

moaz-mokhtar:
I did run `cargo test` and found differences. Check commit: https://github.com/tijlleenders/ZinZen-scheduler/pull/353/commits/91c2997932f4bbb28f30822958523010bbba1184

moaz-mokhtar:
> 



> It is simple. If I say 'Run 5km weekends' do you expect to run 5km or 10km? A daily repeat on a timeline filtered for Sat + Sun would result in 2 steps (2x5km) - a weekly repeat would result in 1 step (1x 5km with a timeline of Sat + Sun).

Thanks, I'm going to change

moaz-mokhtar:
I found many many cases that contains `on_days` with `daily`, so should I change all of them or just who changed in this a prev commit?

moaz-mokhtar:
> I found many many cases that contains `on_days` with `daily`, so should I change all of them or just who changed in this a prev commit?

As expected, system behavior is not as logical as it should be. Below is an example until I test all affected cases:
- For test case `before_7`, it should schedule a task `run` on Sats and Suns for a month timeline. So:
   - when set `"repeat": "weekly"`, it schedule it on a single day only which `Sat`.
      -  I did test swap `on_days` and the same result only `Sat`s.
   - when set  `"repeat": "daily"`, it schedule it correctly of `Sat`s and `Sun`s

moaz-mokhtar:
> > I found many many cases that contains `on_days` with `daily`, so should I change all of them or just who changed in this a prev commit?
> 
> As expected, system behavior is not as logical as it should be. Below is an example until I test all affected cases:
> 
>     * For test case `before_7`, it should schedule a task `run` on Sats and Suns for a month timeline. So:
>       
>       * when set `"repeat": "weekly"`, it schedule it on a single day only which `Sat`.
>         
>         * I did test swap `on_days` and the same result only `Sat`s.
>       * when set  `"repeat": "daily"`, it schedule it correctly of `Sat`s and `Sun`s

The same concern for the other 3 cases as per commits: https://github.com/tijlleenders/ZinZen-scheduler/pull/353/commits/7c51c6edbf0a866ab5d79b53a6ae4223a383ada2 and https://github.com/tijlleenders/ZinZen-scheduler/pull/353/commits/d4451c424884aa83bcb8168cbb383dcdeb2f8e39

tijlleenders (approved):
Nice - on to the next test case!

Issue #359, opened 2023-07-04
test: fix case `sleep-1`


moaz-mokhtar:
**Describe the bug**
Fix case `sleep-1`

PR #363, opened 2023-07-07
test: fix test case `sleep-1`

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/359

--- 
Fix test case `sleep-1`


tijlleenders:
This is the overflow scenario.  
The steps before applying filter should start iterating at 22:00 on the day before the requeated calendar. So 24h steps from 22-22, ending at 22 on the day after the requested calendar.

Then apply filter soo only steps 22-6 are left. That should fit 8h no problem. 

In the output things are split on day boundaries and anything outside of requested calendar is discarded.

src/models/slots_iterator/iterator.rs
```diff
@@ -16,12 +20,29 @@ impl Iterator for TimeSlotsIterator {
...
-                        //next_start_position is 'on' the current slot
-                        result.slots.insert(Slot {
-                            start: slot.start,
-                            end: next_start_position,
-                        });
+                        // TODO 2023-07-07: Use TimingScenario to avoid wrongly slots for overflow slots
+                        if let Some(filters) = &self.filters {
+                            // Determine the timing scenario based on the `after_time` and `before_time` inputs
+                            let timing_scenario =
+                                determine_timing_scenario(filters.after_time, filters.before_time);
```
tijlleenders, line 29: The iterator should not have to know about `timing_scenario` (after_time or before_time).  

This sleep-1 case is more about adjusting the step generation.  
Concretely, just add an adjustment at line 54 of `step_generator`:  
- subtract a few hours (difference between after and 0:00) from `start`
- add a few hours (different between 0:00 and before) to `deadline`

tijlleenders (changes requested):

src/services/splitters.rs
```diff
@@ -230,6 +230,7 @@ pub fn split_slots_into_1h_slots(slots: Vec<Slot>) -> Vec<Slot> {
 ///         Task `Sleep` 00 - 06 (next day)
 /// ```
 pub fn split_crossed_tasks(tasks: &mut Vec<Task>) {
```
tijlleenders, line 232: @moaz-mokhtar 
Anything referring to `Task` should not be in the `placer` service.

tijlleenders, line 232: @moaz-mokhtar 
Can you rename `split_crossed_tasks` to `split_tasks_on_midnight` - or something similar - so the intention is clear?  
Nobody knows what a 'crossed_task` is just by reading these words.

tijlleenders (changes requested):
Glad this is working!  
There is one regression on impossible-2 that need correcting before this can be merged. It's probably related to the ordering of Steps. The expectation is correct.

src/services/step_generator.rs
```diff
@@ -49,8 +51,30 @@ impl Goal {
...
             return steps;
         }
-        let start = self.start.unwrap_or(calendar_start);
-        let deadline = self.deadline.unwrap_or(calendar_end);
+        let mut start = self.start.unwrap_or(calendar_start);
+        let mut deadline = self.deadline.unwrap_or(calendar_end);
+
+        if let Some(filters) = self.filters.clone() {
+            let timing_scenario =
+                determine_timing_scenario(filters.after_time, filters.before_time);
```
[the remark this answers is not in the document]
moaz-mokhtar: That is correct, it more readable, I will fix this

src/services/step_generator.rs
```diff
@@ -49,8 +51,30 @@ impl Goal {
...
+
+            match timing_scenario {
+                TimingScenario::Overflow => {
+                    dbg!(&start, &deadline);
+                    //- set `start`: subtract a few hours (difference between after and 0:00) from start
+                    let after_time = filters.after_time.unwrap();
+                    let diff = 24 - after_time;
+                    start = start - Duration::hours(diff as i64);
```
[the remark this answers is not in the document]
moaz-mokhtar: Sorry I don't get this

src/services/filter/filter_timing.rs
```diff
@@ -2,42 +2,10 @@ use chrono::{Duration, Timelike};
...
-        (Some(_), None) => TimingScenario::AfterOnly,
-        (None, Some(_)) => TimingScenario::BeforeOnly,
-        (Some(after), Some(before)) if after <= before => TimingScenario::Bounded,
-        (Some(after), Some(before)) if after > before => TimingScenario::Overflow,
-        _ => TimingScenario::Unbounded,
-    }
-}
-
 /// Filtering timeline based on before_time and after_time fields in TimeFilter
 pub(crate) fn filter_timing(
```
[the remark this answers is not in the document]
moaz-mokhtar: I designed this to simplify different scenarios!
if you check the match statement for timing_scenario, you will find each scenario have different way of adding slots except both `TimingScenario::AfterOnly ` and `TimingScenario::BeforeOnly`

src/models/slots_iterator/mod.rs
```diff
@@ -33,8 +33,27 @@ impl TimeSlotsIterator {
...
+        dbg!(&result.timeline);
         result.current_start_position = result.timeline.slots.first().unwrap().start;
         result
     }
 }
+
+/// Enum representing Timing Scenario for the provided
+/// timing range (after_time and before_time)
+#[derive(PartialEq, Debug)]
+pub enum TimingScenario {
```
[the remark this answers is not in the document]
moaz-mokhtar: Yes, I think

src/services/splitters.rs
moaz-mokhtar: I just collected all functions that do some split or similar functionalities in a one module named `splitters`

moaz-mokhtar: Ok

moaz-mokhtar:
> Glad this is working! There is one regression on impossible-2 that need correcting before this can be merged. It's probably related to the ordering of Steps. The expectation is correct.

Thank you for your idea.
Actually it working but for similar issue is not.
For example for the case `bug_215` which resolved before already in a previous PR; because we add extra day before and after, that caused to calculate different flexibility value which leads to different schedule for some goals like Sleep.
So that means, we will consider the different flexibility after adding extra day before and after.

tijlleenders:
@moaz-mokhtar 
Good catch, after this PR let's work on #365 so comparing test cases between PRs will be easier.  

So please correct expectation for bug_215 and i293_postpone_1 - and check if there is still a regression.

src/services/filter/filter_timing.rs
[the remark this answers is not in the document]
tijlleenders: > if you check the match statement for timing_scenario, you will find each scenario have different way of adding slots 
I guess you mean removing slots?

src/services/step_generator.rs
```diff
@@ -49,8 +51,29 @@ impl Goal {
...
+        if let Some(filter) = self.filters.clone() {
+            let timing_scenario = filter.determine_timing_scenario();
+
+            match timing_scenario {
+                TimingScenario::Overflow => {
+                    dbg!(&start, &deadline);
+                    //- set `start`: subtract a few hours (difference between after and 0:00) from start
+                    let after_time = filter.after_time.unwrap();
+                    let diff = 24 - after_time;
+                    start = start - Duration::hours(diff as i64);
```
tijlleenders, line 66:
```suggestion
                    start = start - Duration::hours(24 - filter.after_time.unwrap() as i64);
```
Something like this is still readable - but is one line instead of four - and without a comment.

moaz-mokhtar:
> @moaz-mokhtar Good catch, after this PR let's work on #365 so comparing test cases between PRs will be easier.
> 
> So please correct expectation for bug_215 and i293_postpone_1 - and check if there is still a regression.

Thanks
For PR https://github.com/tijlleenders/ZinZen-scheduler/issues/365, I think it is already implemented as below:
- Run `./scripts/test` to run all tests without debug logs
- Run `./scripts/test TEST_NAME` to run a specific test `TEST_NAME` with debug logs
Command `./scripts/test` can cause debug logs if using `dbg!`. I'm using it while debugging then remove it before merge PR

src/services/splitters.rs
tijlleenders: @moaz-mokhtar That is not a good idea. Just because it does something similar doesn't mean it should be grouped together.

tijlleenders (approved):
Fixing the test case with an IF statement is a short term solution ... but agree that we'll cross that bridge (refactor to hanle filtering any timeline with custom start/end/holes) when we get there.  
For any non-resolved comments I'll make separate issues.

moaz-mokhtar:
> Fixing the test case with an IF statement is a short term solution ... but agree that we'll cross that bridge (refactor to hanle filtering any timeline with custom start/end/holes) when we get there. For any non-resolved comments I'll make separate issues.

yes, that is better. Let is create a dedicated issue for this

Issue #357, opened 2023-07-03
Fix expectation and observed for children-with-over-duration


tijlleenders:
**Describe the bug**
Update expectation and fix observed.

kobe-reygel:
fixed during https://github.com/tijlleenders/ZinZen-scheduler/pull/380

PR #367, opened 2023-07-13
test: fix case children_with_over_duration

moaz-mokhtar:
Resolves https://github.com/tijlleenders/ZinZen-scheduler/issues/357

---

Fix case `children_with_over_duration`


moaz-mokhtar:
For the manual calculation `children_with_over_duration` `v1`, I did create a test case named `children-with-over-duration-reorder-issue` just to show that the `goal.id` have an affect on the ordering of `Step`s when they have the same `Step.flexibility`. That is why in case `children-with-over-duration` it passed if we put step `Write report` as last priority based on it's `goal.id=5` and goal `Project B filler.id = 3`.

tijlleenders:
@moaz-mokhtar 
> until decide to change the manual calcualtion (theory) or change the system behavior.

This has already been decided on 2023-07-20 : "The Rust should follow the steps in the excel exactly ... unless there is a theoretical error of applying the algorithm in the excel." 

We are not changing the theory to fit the code behavior.

PR #374, opened 2023-08-14
fix goals hierarchy: enable splitting of filler goals

kobe-reygel:
goals-hierarchy test case was broken because the filler goal wasn't broken up into steps.
This happened because it was still a 'parent' of the concrete goal, and therefore wasn't considered in the splitting algorithm.

Still debating if the cleaner solution would be to have the 'filler goal' be a child of the original budget goal instead of changing the parent into a 'filler' goal. This would - in my opinion - work better with the philosophy of the algorithm.

With the philosophy of 'no unnecessary changes, take the simplest working solution' I decided to keep the current implementation and fix this detail. Can be refactored later.

Moved goals-hierarchy to stable testset and manually validated it against the corresponding excel for correctness. All other tests are still passing.


tijlleenders (approved):
Nicely done @kobe-reygel 

I think this is the right approach for now - as generating a new GoalId for the filler Goal would then also add extra work in the output to 'connect' the filler Tasks back to the filler parent GoalId. Alternative would be to auto-generate filler goals in the UI ... which probably is confusing and in caae of repeating goals impossible?

PR #376, opened 2023-08-24
Treat min budget as steps

kobe-reygel:
* removed the broken BudgetMinWaitingForAdjustment strategy
* opted for a more simple solution to get the baseline working faster: treat 'min budgets' as generating the right amount of steps to satisfy the min budget
* this should fix testcase budget_with_no_children. Unfortunately, there is a bug in the conflict detector: the wrong slots are chosen in the step_placer regarding the 'choose the least conflicting slot' rule
* Already merging this PR because I manually checked that replacing the budget of 'min 40h work per week' with a goal 'work exactly 40h per week' yields exactly the same result (and thus suffer the same wrong placement because of the conflict bug). Therefor I am confident the budget part works as intended.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,6 @@ pub fn run_scheduler(input: Input) -> FinalTasks {
     let steps = generate_steps_to_place(input);
     log::debug!("{:#?}", &steps);
     let placed_steps = step_placer(steps);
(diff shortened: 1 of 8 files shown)


tijlleenders (approved):
Allright!
We're moving in the right direction :)

PR #377, opened 2023-08-27
solve bug in conflict resolution to fix budget with no children testcase

kobe-reygel:
fixed budget with no children:
* there was a bug in the conflict resolution phase: there were multiple steps with the same id, making them not count towards conflict detection.
* disabled filtering on id for now: this has as consequence that a step will count towards its own conflicts. This does not change the order of steps with least conflicts.
* Pull request is in draft until the root of the issue (duplicate ids) is solved

Solving this problem also fixed impossible-2 and issue-276-weekdays-filter-on-budget. Moved them into stable testset


kobe-reygel:
Solved the duplicate id's issue by extending the counter mechanism to also be used when splitting steps: now splitted steps should also have unique ids.

Temporary solution of disabling the filtering was reverted. All test cases still work so I think this PR is ready for final review

tijlleenders (approved):
Awesome work!

Agree that including the js features will not impact the wasm build size. This is not called directly and so no bindings need to be generated.  

Excluding it like [this](https://stackoverflow.com/questions/51821451/how-to-use-a-crate-only-for-a-given-platform) would only complicate the code with no real benefit.

Having a counter suffix for Steps of thr same goal origin is an elegant solution - with the benefit of signalling extra information while debugging. Perfect!

PR #378, opened 2023-08-27
add scripts for pprint debug types in VSCode

tijlleenders:
Hey @kobe-reygel 
 
Screenshot 2023-08-27 170954

**Thanks for the pointers!**  
I was able to pprint in VSCode based on the python file provided in [intellij plugin issue](https://github.com/intellij-rust/intellij-rust/issues/10219) combined with a self-created launch config with the initCommands added:  

```
{
    "configurations": [

    {
        "type": "lldb",
        "request": "launch",
        "name": "Cargo test pprint",
        "cargo": {
            "args": [
                "test",
                "--no-run",
                "--lib"
            ]
        },
        "args": [],
        "initCommands": [
            "command source '${workspaceFolder}/my_type_formatters_for_debugging'"
        ]
        
    }
    ]
}
```

1. The question now is how to best structure this into the repo so someone else using VSCode can easily use it - but it won't bug you if you're using intellij or something else. Having the files in the root is not very nice. 
2. Also, my `launch.json` is not well configured to be called when you start tests from the `rust_tests.rs`... 
3. On top of that, the .vscode is currently gitignored. Should we keep it that way? 
4. We can further strip the python code to only print the info we need - so hour precision for time ... and maybe a space between the year and the date (easier on the eyes when there are less dashes).

I could use some help here ... not too familiar with launch configs and the Rust test ecosystem.


kobe-reygel:
I'm unfortunately also very inexperienced when it comes to launch configs.
Next time I debug I'll do it in VSCode, and I'll let you know if this works right out of the box.

If it works right out of the box - and we can put these files in a non-root folder indeed - I don't mind pushing this to the repository. It shouldn't interfere with people not on VSCode. We should be able to push only these specific files to the repo and still gitignore the rest of the .vscode repo. I think this is best left ignored, that leaves people the option to configure their vsCode repository the way they want to.

If it still requires a lot of setup/tinkering, I'd just document the steps thoroughly and not put it in the repository by default.

I'll let you know when I had a chance to test this out! In any case, nice that it worked for you! 🥳

tijlleenders:
Agree with all your comments.  

We can `git add -f launch.conf` to the `.vscode` folder so it's not ignored and move the rest of the required files to a vscode-dev-dependencies directory.  

Probably something similar can be done for intellij so that you don't have to manually do this on every new dev environment.

tijlleenders:
Thanks for the good contributions @moaz-mokhtar !
I wonder if there's a way to let the Rust analyzer quick actions use the script too... but this is quite good already.
image

moaz-mokhtar:
> Thanks for the good contributions @moaz-mokhtar ! I wonder if there's a way to let the Rust analyzer quick actions use the script too... but this is quite good already. image
> 
> We'll work out the BTreeMap some other time.

Thanks @tijlleenders , this script to be used for debugging, how to do you think this will add value to the deployment?

kobe-reygel:
Thanks for your effort @moaz-mokhtar ! I can confirm this works perfectly with little setup on my machine, so I think we can merge this.

I do agree with @tijlleenders that it is unfortunate that this only works when you run the configuration, and not in the context menu. I also tend to run specific tasks with the context menu. I could not (quickly) find a way to enable this. 

This should not be too much of a problem, we can just add configurations for specific tests, but it is unfortunate

kobe-reygel (approved):

moaz-mokhtar:
> Thanks for your effort @moaz-mokhtar ! I can confirm this works perfectly with little setup on my machine, so I think we can merge this.
> 
> I do agree with @tijlleenders that it is unfortunate that this only works when you run the configuration, and not in the context menu. I also tend to run specific tasks with the context menu. I could not (quickly) find a way to enable this.
> 
> This should not be too much of a problem, we can just add configurations for specific tests, but it is unfortunate

Thanks @kobe-reygel  and have a good work

PR #380, opened 2023-09-02
do not generate steps from non-leaf nodes

kobe-reygel:
not generating steps from non-leaf nodes fixes:
demo_2_with_filler_with_budget.

still in draft because

- test case does not 100% correspond with excel, but I think this is correct. Let's discuss this in a separate meeting.
- I want to write sufficient tests before merging


tijlleenders:
Excellent observation @kobe-reygel ! 
Agree with you the Excel is not correct.

tijlleenders:
@kobe-reygel Totally agree the `children_with_over_duration` is also wrong in the Excel! Sharp eye ;)
I've corrected both Excels.

tijlleenders (dismissed):
Looking good!

kobe-reygel:
@tijlleenders I implemented the necessary tests for the 'do not generate steps from leaf node'.

Also good news: all excel test cases now work. I therefor removed and cleaned up the 'broken testset' tests 🥳 

Cleaned up some tests: steps were not checked for equality correctly because there was an equality implementation somewhere that only checked for 1 field. Tests should now be more correct

I think this pull request is now ready to be merged

tijlleenders (approved):
Excellent!
Nice work getting the tests correct and green. 💪🏽

PR #383, opened 2023-09-09
382 handle new version api

kobe-reygel:
the testcase is added to end-to-end test the new api. Its expected is unverified. How do we want to go about verifying it?

in draft because I want to write unit tests for the deserialization.


tijlleenders:
Looks good @kobe-reygel 

I do expect the API to change a few more times in the coming weeks before it stabilizes... 

There's a proposal to separate Goal objects from Budget objects to make the UX/UI a little smoother: https://github.com/tijlleenders/ZinZen/issues/1363#issuecomment-1712430805

This will impact the API.  
Maybe we can schedule something to discuss this together with @Tushar-4781

tijlleenders (dismissed):

tijlleenders:
@kobe-reygel For the e2e testcase the expected should be based on an Excel... So someone will have to do that - but for now the essential part is that the scheduler doesn't hang.  

We can probably simplify the test to schedule a simple 1h Task for a goal that includes the problematic fields?
This way it's more about testing the API technically than the scheduler logic.

kobe-reygel:
That should not be a problem, it seems like there are good things coming for simplification.

A meeting sounds nice! But don't worry, focus on an intuitive UX first, adjusting the scheduler to match a new UI should be fairly trivial

kobe-reygel:
> @kobe-reygel For the e2e testcase the expected should be based on an Excel... So someone will have to do that - but for now the essential part is that the scheduler doesn't hang.
> 
> We can probably simplify the test to schedule a simple 1h Task for a goal that includes the problematic fields? This way it's more about testing the API technically than the scheduler logic.

you are right! I simplified the testcase :)
I also added deserializing unit tests, this PR should be ready to go now

tijlleenders (approved):
Excellent!

I'll move this over to the frontend repo.

PR #386, opened 2023-09-12
default goals proposal for api discussion

tijlleenders:
@kobe-reygel 

In preparation for below simplification of the UX/UI, the API will need to change too.  
I'm curious to hear your thoughts on this - and on what to do with all the current testcases...

image

### DIFF
--- a/tests/jsons/api-2023-09/default-goals/input.json
+++ b/tests/jsons/api-2023-09/default-goals/input.json
@@ -1,0 +1,326 @@
+{
+    "startDate": "2022-10-10T00:00:00",
+    "endDate": "2022-10-17T00:00:00",
+    "goals": [
+        {
+            "id": "1",
+            "title": "Daily routine",
+            "children": [
+                "2",
+                "3",
+                "4",
+                "5",
+                "6",
+                "7"
+            ]
+        },
+        {
+            "id": "2",
+            "title": "Sleep 😴🌙",
+            "budget": [
+                {
+                    "on_days": [
+                        "Mon",
+                        "Tue",
+                        "Wed",
+                        "Thu",
+                        "Fri",
+                        "Sat",
+                        "Sun"
+                    ],
+                    "after_time": 22,
+                    "before_time": 8,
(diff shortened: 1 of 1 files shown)


kobe-reygel:
I don't quite fully understand the direction the new UX/UI is trying to go in. Can you show me some full mockups? Alternatively we could schedule a meeting going over the new UI :)

As far as the change in the API goes: I would not redesign the algorithm to fit this new UI (and I don't think that is what you intend either?).

So the internal structs should stay the same imo, only the API/ deserializing step should change. This should not be too hard, assuming the new API's elements can be mapped 1-to-1 on existing concepts in the algorithm.

As far as changing tests goes: unfortunately to be correct I do think that the tests should be re-written to a new API format. This can't be avoided when redesigning the API. The Expected and Output should not change, so it should not be too much work.

One thought I have on this is that it might be better to have what we now call end-to-end tests at the level of the scheduler (with the deserialized rust objects). We would then only need a couple 'integration tests' that are set up as the current e2e tests. They would purely serve as a test for correct (de)serializing and to test that every part of the application works together well. I don't know if this is worth it at this point, but this would leave us more flexible in API redesign.

tijlleenders:
Yes let's schedule a refinement so we can discuss the UX/UI proposal three-ways.  

For remapping internal structs I agree it should be doable. I'll see if I can rewrite some of the existing passing tests. The rest should be abandoned in my opinion - as I think it's best to focus on 'real life' examples instead of contrived edge cases.  

I understand your proposal. It's more efficient (as correct deserialization tests are fewer in number, more unit-test-like, and easier to reason about) .... **if** the API changes frequently.  
Since I don't expect the API to change significantly once we've found a good front-end way of doing things I think we're good the way it is, especially if we drop the other tests. A quick search replace will suffice for those - and like you said, the output thankfully remains the same.

tijlleenders:
We can continue using the current API for now - and make a split to parse an element in the list to a budget or a goal on the Rust side of things.

PR #389, opened 2023-09-15
implement proposal to allow cargo publish

kobe-reygel:


tijlleenders (dismissed):
@kobe-reygel 

Your proposal is sooo much better!!!  
Reading back my proposal really feels like bending backwards in a weird and uncomfortable position.  

```Published zinzen v0.2.0 at registry `crates-io``` 🥳

tijlleenders (approved):

PR #390, opened 2023-09-16
Documentation test generation feature flag

kobe-reygel:
- documented the recently created need for using the generate-tests feature flag
- introduced the concept of ADRs (Architectural Decision Records)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,9 +17,9 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Build
-      run: cargo build --features generate-tests --verbose
+      run: cargo build --verbose
     - name: Run tests
-      run: cargo test --features generate-tests --verbose
+      run: cargo test --verbose
     - name: Install latest nightly
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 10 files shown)


tijlleenders:
@kobe-reygel 
This is a great idea!  

About the feature:  
Wouldn't it be better to have the feature on by default and only disable it manually when doing a ```cargo publish```?  
This way new people on the project will not run into trouble, and the few times we do need to publish and forget the flag we'll get the error message reminding us about the flag required to publish.

kobe-reygel:
Yes that would be an option. I don't have a strong opinion either way, so I'll leave that up to you what you find most convenient :)

Only small counter-argument that I could imagine is that there really is no need to do the work of analysing the JSONS directory every build. Only when we add tests (or delete them). This notion is the reason I opted for the current way of doing things.
This takes literal milliseconds though, so it's not a very strong argument.

Shall I change the working of the feature flag (and the documentation) to the inverse of what it is now?

tijlleenders:
@kobe-reygel 

Yes, OK if you can change the flag and the corresponding ADR we can merge this.

kobe-reygel:
@tijlleenders implemented proposed change and documented it.

tested it with a dry run, the feature flag does what it's supposed to do

tijlleenders (approved):
Great!

Issue #301, opened 2023-04-29
Benchmark fails


tijlleenders:
**Describe the bug**
When attempting to run the benchmark previously developed - it fails.

**To Reproduce**
```
tijl@TFlap:~/ZinZen-scheduler$ cargo run --bin scheduler
   Compiling zinzen v0.1.1 (/home/tijl/ZinZen-scheduler)
    Finished dev [unoptimized + debuginfo] target(s) in 2.05s
     Running `target/debug/scheduler`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: sequence, expected a map", line: 4, column: 16)', src/bin/main.rs:12:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected behavior**
Runs the benchmark previously developed.

**Desktop (please complete the following information):**
 - OS: linux

**Additional context**
Tests that need to be adapted are in tests/jsons/benchmark-tests - or just point the main at different test inputs.

kobe-reygel:
I'll look into this

kobe-reygel:
https://github.com/tijlleenders/ZinZen-scheduler/pull/391 should fix flamegraph to run again.

Both binaries (flamegraph and test-summary) now work again.

test-summary does have a bug: rfind only finds **the last** matching string. this means that only the last test module is printed out as a summary.
@tijlleenders does anyone use this binary? I see 2 possible paths
1) we throw away the binary, just running cargo test works just as well. All unnecessary logging has been removed, so the output of cargo test is relevant (apart from the fact that there is an 'OK' line per test, so the output is rather long)
2) I replace the rfind by something more suitable to print **all the summaries** on top of eachother, instead of now only the last module summary

Personally I tend towards option 1). option 2 seems like quite some work for something that is of minor benefit.

tijlleenders:
Agree with option 1.

kobe-reygel:
@tijlleenders option 1 implemented in [e7b2d9f](https://github.com/tijlleenders/ZinZen-scheduler/pull/391/commits/e7b2d9f916f93eb15d7b72b8fa27a30102f5bd94)

PR #391, opened 2023-09-16
fix flamegraph

kobe-reygel:
fix for being able to run flamegraph again (issue: https://github.com/tijlleenders/ZinZen-scheduler/issues/301)

@tijlleenders flamegraph-rs does not work on MacOS (see documentation in PR), could you run this to verify this allows you to run the benchmarks again with "cargo flamegraph --bin flamegraph"

I picked a random test to supply to the script, I don't know which tests you had in mind to use for profiling?


kobe-reygel:
in [2865563](https://github.com/tijlleenders/ZinZen-scheduler/pull/391/commits/2865563e43b80b2542d1fe37dfbda1b3d51311de) I also removed the old benchmark input and code: it was no longer in use anyway because
1) test generation already ignored the input folder
2) the benchmark code already pointed to different tests

tijlleenders:
Yeah, this works.  
It creates a `flamegraph.svg` and a `perf.data` in the root folder: 
image

image

It would be nicer if we output it to a separate folder and .gitignore the folder.

kobe-reygel:
@tijlleenders I changed the default test to 'demo-2-with-filler-with-budget', I think this is what you mean with 'the default goals test'?

in regards to the output folder: that is something that the user of the command 'cargo flamegraph' has to supply, we can't really define this for the end user of cargo flamegraph. I think we just need to never check in the flamegraph.svg file (I removed the currently checked in file)

tijlleenders (approved):
We don't have a good default goals yet so you're right, the `demo-2-with-filler-with-budget` is probably the closest thing to it.  
I'll be working on a 'complete' default goals in #386.  
The idea is to have the app startup with a whole set of pre-filled goals that corresponds to what most people would input approximately, so that you get an example that you then just have to tweak to your liking - or delete in favor of some goal-tree you build up yourself from scratch.

PR #393, opened 2023-09-16
remove splitting in 1h blocks in favor of counting the overlap

### LINKED ISSUE #210: Performance analysis
Run `cargo flamegraph --bin flamegraph-bin` and check if there is anything REALLY wrong.

Analyze scheduler performance and find hot loops to improve - but **only if user experience is impacted enough!**
As long as the scheduler can run a test case in less than 100ms there is no issue.  

flamegraph

kobe-reygel:
Resolves #210

### DIFF
--- a/src/models/slot/impls.rs
+++ b/src/models/slot/impls.rs
@@ -84,8 +84,13 @@ impl Slot {
         (other.start >= self.start) && (other.end <= self.end)
     }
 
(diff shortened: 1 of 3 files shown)


tijlleenders (approved):
Awesome!!!  

This took the standard tests from **23**s to **0**.83s on my machine.

kobe-reygel:
@tijlleenders yeah the speedup is truly awesome for a fairly small adjustment 🥳

Long live profilers

PR #395, opened 2023-09-22
Resolve issue #388

### LINKED ISSUE #388: Add max per day budget check
**Is your feature request related to a problem? Please describe.**
The max per day is currently not respected.

**Describe the solution you'd like**
Make the scheduler respect the max per day setting.

**Describe alternatives you've considered**
None.

**Additional context**
Example which schedules all 5 hours on the first day - where I expect max 1h per day:  
```
{
    "startDate": "2023-09-15T00:00:00",
    "endDate": "2023-09-22T00:00:00",
    "goals": {
        "1c955b93-729f-4dd9-b334-950fb3731622": {
            "id": "1c955b93-729f-4dd9-b334-950fb3731622",
            "title": "Walk 🚶🏽",
            "filters": {
                "after_time": 14,
                "before_time": 20,
                "on_days": [
                    "Mon",
                    "Tue",
                    "Wed",
                    "Thu",
                    "Fri"
                ]
            },
            "created_at": "2023-09-15T04:38:57.394Z",
            "min_duration": 5,
            "start": "2023-09-15T00:00:00",
            "repeat": "weekly",
            "budgets": [
                {
                    "budget_type": "Daily",
                    "min": 1,
                    "max": 1
                },
                {
                    "budget_type": "Weekly",
                    "min": 5,
                    "max": 5
                }
            ]
        }
    }
}
```

SFBdragon:
Resolves #388 

This commit includes a new test with the input.json, expected.json, and observed.json

Please let me know if I'm on the right track.


tijlleenders:
@SFBdragon Yes this is exactly the `expected.json` I expect.  
Next step is to see where the current scheduler goes wrong:  
- does it setup the right Budget object with five daily budgets and one weekly budget?
- does it setup the right (number of) Steps? 
I'd expect five. Normally this would be one step for a block of 5h (since <= 8h it's scheduled as a block) - but since the max/day is one hour this limits the Step size to 1h ; forcing it to be broken down into smaller Steps.

SFBdragon:
The first apparent issue with the handling of the budgets is here

https://github.com/tijlleenders/ZinZen-scheduler/blob/7c31343c1da3b891dcb903260764f8fe0fc18ab4/src/models/budget/impls.rs#L67

I recommend placing a breakpoint on line 75 and running the i388 test with the debugger, if you'd like to verify.

The issue is that within
- `generate_steps_to_place`
- `configure_budgets`
- `insert_budgeted_goals`
- `insert_goal`
- For each budget in the goal, the code constructs a new `StepBudget` and then on line 75, the step budget is associated with the goal's ID. Notice that if the loop runs twice (which it does for a goal with two explicit budgets; not inherited), the first entry into the `StepBudgets` map is overwritten.

There's a couple options here:
- `StepBudget` could perhaps hold both a daily and a weekly budget 
- The map could associate each goal ID with a vec of `StepBudget`

Currently, this code only deals with budgets directly on the goal node itself. The code that associates the parent budgets with the node doesn't occur until afterwards.

As far as I can tell, we could reduce all of the hierarchical information into a single set of budgets (e.g. daily and weekly) for each leaf node here instead. Or we can just leave all of this up to the step generator, as the code currently does. I'm not under the impression the step generator handles this case correctly, but let's take this one step at a time.

I'm also concerned about how you plan on changing the approach regarding budgets in the future, so please advise how I should move forward, thanks.

SFBdragon:
This fix comes with a few changes in behaviour. 

Most notably, 
- the `find_best_slots_for_first_step` now returns the full array of possible slots, best to worst, such that if the placer determines that the best slot is not compatible with the step's budgets, it can try the other slots as well.
- `min_duration` for goals is now used, rather than one of the budget's minimum duration requirements.

Rust tests for these have been updated accordingly to get them passing.

A number of E2E tests were broken as a result of these changes:
```
e2e::budget_with_no_children
e2e::children_with_over_duration
e2e::i287_mix_budgets_many_goals
e2e::issue_276_weekdays_filter_on_budget
```
These mostly appear to be the results of changed scheduling behaviour, rather than new problems, but I'll leave the confirmation and resolution of these up to the maintainers.

Known outstanding issues with this fix:
- max_duration for goals is now vestigial, considerations should be made to repurpose it or remove it
- the TODO in budgeting.rs line 135 purportedly related to #358 is likely now outdated, I recommend reviewing its relevance as a result of the surrounding changes
- flexibility calculations are broken, as they don't account for budget conflicts excluding slots
- steps are broken up according to the most restrictive block, but this will fail for the following restrictions
    - min_duration: 20
    - repetition weekly
    - filter out weekends
    - daily min 3
    - daily max 5
- in a very busy schedule, it would be best to have steps be closer to the minimum requirement to make busy days possible and taking more time on other days within the maximum budget requirement
- filters are entirely ignored, which may also cause similar issues to the above

This'll be the last of my activity on the project for the time being. All the best for the project going forward!

tijlleenders:
Thanks for the nice wrapup @SFBdragon !
We'll look into it soon.

Good luck with your exams!

tijlleenders:
> the find_best_slots_for_first_step now returns the full array of possible slots, best to worst, such that if the placer determines that the best slot is not compatible with the step's budgets, it can try the other slots as well.  

Agree!

> min_duration for goals is now used, rather than one of the budget's minimum duration requirements.

This corresponds to the changes in `budgeting.rs` function `generate_steps` - agree!

As for the new failing test cases:  
- `e2e::children_with_over_duration` 
  `expected.json` was erroneous (didn't match the Excel expectation) - the new `observed.json` is correct.  
- e2e::budget_with_no_children
  I would expect there to be 40 steps generated for the goal/budget 'Work' - but there are none. 
  Whenever a budget goal doesn't have any subGoals (with a duration) the fix is to add the steps with the name of the budget goal - respecting the splitting business rules. In this case there is no min per day budget, and the number of hours of the min budget per week is 40, so since 40 > 8 that would result in 40 Steps of 1h, each with the Timeline of entire week, as there are no filters on the budget goal.  
  This should be implemented to make the test pass again.  
  Also, the `expected.json` is not identical to the Excel and needs to be corrected.

Then there are these two:  
- e2e::i287_mix_budgets_many_goals
  This one is strange. I don't see a new `observed.json` being generated - but the test is failing for `cargo t`.  
  If I look at the expected and observed - there is no diff...
- e2e::issue_276_weekdays_filter_on_budget
  This one is strange. I don't see a new `observed.json` being generated - but the test is failing for `cargo t`.  
    If I look at the expected and observed - there is no diff...
image


**All in all an excellent PR @SFBdragon - thanks for putting in the time to the elaborate and document all the issues you see !**  

If I have to make one comment it would that be that a monolithic commit is harder to review. If you separate out each logical change hen we can read the individual commits as a story - instead of having to go through your comment and match/map pieces of your comment to the relevant changes in the code. Subsequent comments/changes/discussions are then also centered around the logical changes - making it easier for you too.  
This way of working has the added advantage of also allowing selective pruning of commits - so your PR can get merged faster in case only a single logical piece is not required.

We'll go through the outstanding issues (not related to this PR) at a later point - and make issues / test cases where relevant.

kobe-reygel:
@tijlleenders good news: children_with_over_duration seems correct (but wrong in excel)

in excel, cell EF461, we say that the budget of the week is full. When I count the already scheduled hours (=SUM(D421:D460)) we reach 42 scheduled hours, while the max weekly budget is 45h.

The test case handles this correctly (thanks to the fixes of @SFBdragon and @vinaybadgujar102) by scheduling 3 more hours in the first week in comparison to the excel.

I adjusted the expected.json to reflect this in commit [f62546b](https://github.com/tijlleenders/ZinZen-scheduler/pull/395/commits/f62546bf4ce823751ea236508b8f71319d4335fc). We do need to reflect this change in the excel at some point if you agree.

tijlleenders:
@kobe-reygel The 3 extra hours should be optional, after all required hours are scheduled. I'm not sure if that's taken into account, and we might use the 3 optional hours instead of giving them to a step that actually requires them?... but I would need to devise a test case to prove this.
For now let's go with it.

kobe-reygel:
@tijlleenders that should already be covered by the ordering: filler goals have least priority.

The main thing here is that the test case has a max budget of 45, but in the excel it cuts off after 42, so it seems like a simple counting error while designing the test case.

To be sure I changed the input to have a max budget of 42 (instead of 45) and then we get the exact behavior described in the excel.


So what would you prefer? keeping the input at 45 and changing the expected and the excel to match that
OR
changing the input to 42

tijlleenders:
@kobe-reygel Let's keep it simple and adjust the test case to 42, matching the Excel. The optional diff between max and min budget is something we can add on later.

tijlleenders:
When in doubt, go with the answer "42"

kobe-reygel:
@tijlleenders will do,
I do however not understand what you're getting at with the optional diff - that is not entirely relevant here. There is merely a small error in the excel where there is a mismatch between the written budget of 45 and an actually enforced budget of 42.

This can be solved by either reflecting the written budget to say 42 or reflecting the enforced budget to use 45. The optional diff does not come into play in this case.

This is confusing to type out - we'll go over the excel in a next meeting 😄

kobe-reygel:
> When in doubt, go with the answer "42"

haha good one! @tijlleenders

kobe-reygel:
@tijlleenders in commit https://github.com/tijlleenders/ZinZen-scheduler/pull/395/commits/d3cc2e4fa94c7bc6104cf719663564b6d07a7cf8 I put back the option for generating steps based on step_budgets that are not necessarily attached to goals. This behavior got erroneously removed in earlier commits for this PR.

This solves 2 errorring tests while still keeping all other tests green.

I do not think this is the cleanest solution, but since we're about to take another look at the architecure I think it's best to keep it like this for the moment, so that at least the tests are green again and we can start refactoring/rethinking the architecture.

1 failing test left before this can be merged

tijlleenders:
@kobe-reygel  
Yes, agree this behavior needs to be maintained.  

As per our discussions a quick way to differentiate between a goal and a budget (given we consider this the same type & technical structure) is whether there is a goal.min_duration defined.  
- If goal.min_duration is None && budget fields is Some (ie day of week filter) defined, **it is a budget.** 
- if goal.min_duration, **it's a goal.**  
- if goal.min_duration is None && budget fields is None it's just a placeholder for hierarchical structure that could possibly be filtered out before sending to the scheduler.


A budget always has a minimum that *has* to be placed each week, so it always needs to generate at least the minimum steps that contend with regular goals for slots.

tijlleenders:
@kobe-reygel 
As for the remaining testcase I think your red block comment made 2023-09-07 in the Excel for `i287_mix_budgets_many_goals` nails down the problem: the current algorithm is not using 1h Steps when translating a Budget into a (set of) Steps.  

It's probably not super efficient - but naïvely restricting Budgets to _only_ be allowed to be converted into Steps of max 1h (as is done in the excel) would probably solve this issue.  
There is no reason to assume a minimum block size for Steps that are a result of a Budget. The min in the budget refers to the min constraint taken over a whole day or week (possibly in separate Steps => Tasks), not a minimum length/duration of a single scheduled Step/Task. So what is important in the generation of Steps from Budgets, is that for a given day, if it has a min budget, that there are _exactly that number_ of 1h Steps in total that also have a Timeline restricted to that day.

Hopefully doing that will not create a cascading effect on other test-cases ; otherwise I'd suggest doing the pair-programming session with @thinkrapido first - and possibly do some architectural/structural changes.  
When it's too hard to make changes - change the structure first - so that it _then_ becomes easy to make the change.

kobe-reygel:
@thinkrapido @tijlleenders I got the tests to succeed for this PR. I noticed the following bug: https://github.com/tijlleenders/ZinZen-scheduler/issues/406 
but decided to not tackle it in this PR, so this can finally be closed. 

I'm not entirely convinced of the approach, but let's get that cleared up in context of the bigger architectural refactorings that are planned :)

tijlleenders (approved):
Yeah, good eye.

A Budget definition received from UI/input should always have the following:
- min-max / day
- min-max / week
- on_days filter
- after-before filter

How this is split into Steps and Budget objects in Rust is up to us.

thinkrapido:
Nice, now we can start from a clean slate. Cheers :+1:

PR #401, opened 2023-11-01
refactor: improved readability

thinkrapido:


tijlleenders:
@thinkrapido 
Just curious: why does improved readabilty require nightly? 

Is it the 'with-logging' feature?  
Where can I find docs for that? I'd like to understand what it does.

thinkrapido:
> @thinkrapido Just curious: why does improved readabilty require nightly?
> 
> Is it the 'with-logging' feature? Where can I find docs for that? I'd like to understand what it does.

https://github.com/rust-lang/rustfmt

blank_lines_lower_bound for instance is only on nightly toolchain recognized

.editorconfig
```diff
@@ -0,0 +1,4 @@
+root = true
+
+[*]
+trim_trainling_whitespace = true
```
kobe-reygel, line 4: Should be trim_trailing_whitespace I suppose?

kobe-reygel:
@tijlleenders our rust build for the tests in the pipeline also already runs on nightly. The logic behind this (if I understand it correctly) is catching things early: preventing us from merging stuff that will be regarded as 'wrong' in the next stable build.

I therefore think that Romeo's suggestion of also using it for rustfmt is a good one

Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["algorithms", "date-and-time", "wasm"]
 [features]
 experimental-testset = []
 skip-test-generation = []
+with-logging = []
```
kobe-reygel, line 16: Great idea

kobe-reygel (approved):
Looks good

PR #404, opened 2023-11-04
Rom/ /clean code

thinkrapido:


tijlleenders (approved):
Thanks @thinkrapido 

BTW you can set your IDE to `cargo fmt` on save so it doesn't end up in separate commits.

PR #405, opened 2023-11-04
Rom/ /date normalization

thinkrapido:


tijlleenders (approved):
Good stuff @thinkrapido !

Please keep in mind that at some point we'll want to increase the Slot granularity to 15/5/1 minutes or even seconds.

PR #415, opened 2023-11-12
Tijl/ /validate experimental test cases

tijlleenders:
@kobe-reygel 
I went through the remaining experimental testcases and validated them - often backed by an Excel.  
I removed some irrelevant/unclear/large-for-nothing testcases.
  
The Budgets are in the new format - so they will break. That's why I left them in /validated.

`sleep-1` and `weekly-1` should pass but there is some whitespace that makes the json diff fail... Logged #417 for that.


kobe-reygel:
@tijlleenders fantastic work, that must have been a massive effort

I double checked those failing testcases, they fail because of the new input format (and a corresponding runtime panic), not because of whitespace.
I documented this in the comments of https://github.com/tijlleenders/ZinZen-scheduler/issues/417 and closed the issue

I made https://github.com/tijlleenders/ZinZen-scheduler/issues/422  instead
I'm not sure if making this a bug is the way to go, or if we should call this a new story to support the new input format. But regardless now there is a paper trail of why those tests fail and an issue to fix it :D



The 'test-to-avoid-failing-build-when-experimental-tests-dir-is-empty' is pretty ugly, I'll see if I can do something about that without losing too much time, and then this can be merged

kobe-reygel (approved):
see comments, looks good!

PR #416, opened 2023-11-12
add branching convention to docs

### LINKED ISSUE #413: naming conventions for development on branches
I'd like to suggest a naming convention for branches we are working on.

[developer shortcut]/[issue number]/[meaningful name]

like: rom/411/new-scheduler

The branches should only be developed by the distinct developer and not mixed up.
This helps retrace the development train of each developer and does not get confusing.

Finished branches should be merged into the main branch after reviewing the pull request and then deleted.

If someone likes work on a branch of another developer, then branch from this one with the above naming convetion.

tijlleenders:
Resolves #413

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,15 @@
+# Branching convention
+[developer shortcut]/[issue number]/[meaningful name]
+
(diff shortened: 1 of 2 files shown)


kobe-reygel (approved):
sounds good, clearly documented

PR #420, opened 2023-11-14
added cargo.lock to repo

thinkrapido:
It doesn't make sense to exclude Cargo.lock from repo


tijlleenders:
> It doesn't make sense to exclude Cargo.lock from repo

@thinkrapido Please use the issue templates - or at least motivate your request.  

Your request is valid according to guidance on [crates.io docs](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) :  
"If you're building an end-product which are executable [...] check Cargo.lock into git."  

More info : [Why do binaries have Cargo.lock in version control, but not libraries?](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)

tijlleenders (approved):

PR #431, opened 2024-01-06
Tijl/ /rc calendar

tijlleenders:
After review from @thinkrapido this implementation and the abstractions seems to fit the need.  

The approach is to use a main calendar of Vec<Rc<Hour>>.  
Each Hour on the Calendar can be Free or Occupied.  

Each Goal (simple or budget) gets translated to one or more Activities by activity_generator.rs.  

Each Activity has an internal calendar_overlay of Vec<Option<Weak<Hour>>>.  
This get filtered on Activity creation for any Hour that doesn't fit the Goal constraints.  

After that, the activity_placer.rs doesn't know anything about dates/times - it just looks at how many blocks are available, and schedules each Activity in turn, counting conflicts via Rc::weak_count, updating the Calendar, which auto-updates the Activities overlay via the Rc:Weak that get invalidated when an Hour is converted from Free to Occupied.


tijlleenders:
Oh I see I still have some Clippy things to fix...

tijlleenders:
@thinkrapido @kobe-reygel 

Hey people,
Just an update on the rc-scheduler approach.

**The min/max day & week budgets are working fine now!**

0.  The only reason the `default_budgets` test is failing is because I don't like the extra sleep hours (the diff between min and max) being scheduled from 22-23, with a gap from 23-0. This could be fixed easily by inverting the 'walk' of finding the best_hour_index - starting from the back to the front whenever the 'after' time_filter is greater than the 'before'.

1.  I've been quite lazy in the code. There is a lot of cleaning up to do.  
Some thing are duplicated on purpose so I don't have to keep two contexts in my mind when walking through the code. I don't think it is worth factoring out with an extra argument for the context ... but there are still some opportunities for factoring out parts of functions that don't depend on the context they are called in - or moving functions to the 'appropriate' module.

2.  I've ignored clippy so far. That's another thing.

3.  Finally, there's a bunch of TODO's I sprinkled all over the place whenever I thought of something while working on something else.


I'll not have time to code till Friday morning - as going to the office physically consumes the little 'free/coding/ZinZen' hours I have in a day - so feel free to do stuff - if you feel like it.

vinaybadgujar102 (approved):

PR #433, opened 2024-01-23
add test budget-with-subgoal

tijlleenders:
Sorry for the messy commit history. 

I didn't want to loose out on the improvements from the other PRs waiting for review - and then I tried rebasing for the first time :)  This PR's commits start at [fix unwrap bug](https://github.com/tijlleenders/ZinZen-scheduler/pull/433/commits/682e2e2c0d0569b5ff6242cd0c43e6be7d38e6bf) - a commit that wasn't necessary and is overturned later... Still haven't figured out how to clean that up so I get a commit history that reads like a story...

In terms of the strategies outlined below, the simple Goals get scheduled first anyway - so there was no special change needed for current test cases. Just filtering the Hours based on the Budget and updating the Budget when placing the simple Goal.

Basically what I did to fix this issue:
- Treat a Goal that has children as valid for generating Activities from (don't skip if children present, like before)
- Updated the Budget struct so it also stores the appropriate Filters. This can then be used in generating the calendar_overlay for the simple Goal activities - by checking if the Simple Goal is part of any Budget - and if it is - use the Filters upon initial construction of its overlay. We don't want that to happen more than once. It's a static constraint - not a dynamic one like Budgets.
- Updated the dynamic constraint in the placer loop ( `activity.update_overlay_with(&calendar.budgets);` ) to also update Simple Goal overlays when the Simple Goal is part of a Budget
- Updated placer to update the Budget if a Simple Goal is part of a Budget

=============================
Original PR/issue description:  
Test where someone has a workbudget of 40h on weekdays. Below that, as a subgoal, we add a 1h 'Plan my work week'.  

Since the goal is more specific than the generic budget - it should be planned first - and also reduce the budget by one hour, as it counts as work (since it's a subgoal of work).

Technically, I see a few ways of approaching this:  
1. make an activity for 'Plan my work week' - possibly with new type and somehow integrate that in the scheduling so that it's always earlier than the budget it is part of... sounds complicated? 
2. Or if we place it in the same `Vec<Activity>` as the budget - but at the beginning of the Vec - it will get scheduled first - as it will always have the same flexibility as the other budget Activity hours?
3. Keep the 'Plan my work week' on a separate list, and replace/switch out the first Task hour of 'Work' budget Tasks (after they have been scheduled normally taking 40h of the week) with the simple goal title.


tijlleenders:
@thinkrapido @kobe-reygel  
This is a failing test case. I've not made a separate issue for it - but have added the PR to the project board parking lot - where it appears just like an issue - and can be self-assigned if you'd like to work on solving this.

src/models/activity.rs
```diff
@@ -284,9 +286,18 @@ impl Activity {
     ) -> Vec<Activity> {
         let mut activities: Vec<Activity> = vec![];
 
-        let compatible_hours_overlay =
-            Self::get_compatible_overlay_hours(goal_to_use, calendar, time_budget);
-
+        let compatible_hours_overlay = Activity::get_compatible_hours_overlay(
```
thinkrapido, line 289: Why is this code refactored to code duplication?

src/models/activity.rs
```diff
@@ -216,7 +218,7 @@ impl Activity {
         goal: &Goal,
         calendar: &Calendar,
     ) -> Vec<Activity> {
-        if goal.children.is_some() || goal.filters.as_ref().is_none() {
+        if goal.filters.as_ref().is_none() {
```
thinkrapido, line 221: This is good. I also removed goal.children.is_some() in the filler goal test case. Therefore ok.

src/models/calendar.rs
```diff
@@ -244,6 +244,7 @@ impl Calendar {
                             originating_goal_id: budget_id.clone(),
                             participating_goals: descendants_added,
                             time_budgets: get_time_budgets_from(self, goal),
+                            time_filters: goal.filters.clone().unwrap(),
```
thinkrapido, line 247: remove duplication

src/models/calendar.rs
```diff
@@ -258,6 +259,7 @@ impl Calendar {
...
```
thinkrapido, line 262: remove duplication

src/models/calendar.rs
thinkrapido: see below

src/models/calendar.rs
thinkrapido: see above

thinkrapido (commented):
Mainly resolve some code duplications.

src/models/activity.rs
tijlleenders: See https://github.com/tijlleenders/ZinZen-scheduler/pull/446

src/models/calendar.rs
tijlleenders: @thinkrapido Please explain what the duplication is.  
`time_budgets` is a completely different thing from `time_filters`.

src/models/calendar.rs
tijlleenders: see below

Tushar-4781 (approved):

PR #434, opened 2024-01-23
add test filler-goal

tijlleenders:
Test for filler-goal.  

The idea is that when you have a big one-time thing of several hours - you might put a reservation of X hours on the calendar. 
Then as you think of more things - you add subGoals that can be independently achieved. If there is time left over in the parent goal after the specific subGoals have been achieved - that time also gets reserved on the calendar. 

Another example would be a vacation day - where you have a few things you want to do - so you make specific subGoals for that - but the rest of the day is left open - but you still want the rest scheduled on the calendar as 'vacation day'.


tijlleenders:
@kobe-reygel @thinkrapido Another failing test case someone can pick up.

thinkrapido:
done by https://github.com/tijlleenders/ZinZen-scheduler/pull/447

tijlleenders:
@thinkrapido 
I'm sorry - I didn't provide enough detail to make it clear what is required to solve this issue. Thanks for the work and solution provided. I'll see if I can build on that.

moaz-mokhtar:
Hello @kobe-reygel @thinkrapido 

I hope you are doing well.

I will work on this

moaz-mokhtar:
I want make sure that I understand the issue to be resolved.
- I did run tests on branch `tijl/-/add-filler-goal-test` and found that `filler-goal` failed which showing different results from expected.
- What I understand that here we want to allow sub goal which exceed parent goal's deadline to be reserved in the calendar.
- I found PR #447 which seems related to `filler-goal` itself, I think I need to create new test case for this PR specifically if I'm not wrong!!!

tijlleenders:
@moaz-mokhtar Why do you think the subGoal needs to go beyond the parent goal's deadline?  

The subGoal is >>part of work Goal<< and as such the Activity it generates:
- should at least satisfy the time conatrainta of the parent. 
- should be linked to the Budget which is also related to the Activity 'Work' generated by the parent Goal.

moaz-mokhtar:
I did debug yesterday and would like to share below information.

- Both tests `stable/default-budgets` and  `stable/filler-goal` doesn't generate tasks which have children in addition it is not validating that child is valid. But `stable/default-budgets` success because it is adjusted. I think it should generate task with `Daily habits` if I'm not wrong.

- I'm proposing a long-term solution to be applied when convenient. It is for set of rules that validating input goals:
    - System will not proceed when input goals not validated and will return error with related information
    - The `start` and `deadline` of a child's goal should not fall outside the `start` and `deadline` of its parent goal.
    - The budgeting `budget_config` of a child's goal should not fall outside the budgeting of its parent goal.
    - The filtering `Filters` of a child's goal should not fall outside the filtering of its parent goal.
    - The `min_duration` of a child's goal should not fall outside the duration, budgeting, or filtering of its parent goal.
    - Parent goal shouldn't allow child goal/s to take more duration that causing other child goals to fall outside it.

- Happy scenario solution which I will work on it is will consider input validated and child goals fall inside the parent's goal

- My proposal is:
    - pass `parent_goal` information when generating activities to adjust based on that as below algorithm:
    ```rust
    // goal.rs:42
    pub fn get_adj_start_deadline(&self, calendar: &Calendar, parent_goal: Option<Goal>) -> (NaiveDateTime, NaiveDateTime) {
    // if parent_goal.is_some -> means it is child goal, so do below:
    // adjust generated goals with respect to parent_goal timing constraints
    ```
    - For this stage it will be simple which will not apply previous validation for input goals.

tijlleenders:
@moaz-mokhtar Please focus comments in this PR solely on the filler-goal test case.  
If you think there is something wrong / to be improved in other test cases - please make a separate issue for that. Otherwise it will become very confusing very fast.  

All I'm asking in this PR is a passing test case based on correctly generated Activities and output Tasks - while keeping input Goals immutable.

tijlleenders:
@moaz-mokhtar There have been some improvements to main while this PR was outstanding.  
Please resolve the merge conflicts, respecting the improvements in main.

tijlleenders:
@moaz-mokhtar I think I merged correctly - but that resulted in the tests failing.  
Please double-check and investigate.

src/lib.rs
```diff
@@ -90,20 +91,17 @@ pub fn run_scheduler(
...
     let mut calendar = Calendar::new(start_date, end_date);
-    dbg!(&calendar);
 
     calendar.add_budgets_from(goals);
 
     //generate and place simple goal activities
     let simple_goal_activities =
         activity_generator::generate_simple_goal_activities(&calendar, goals);
-    dbg!(&simple_goal_activities);
+    let simple_goal_activities = adjust_parent_activities(&simple_goal_activities, goals);
```
tijlleenders, line 100: Please make this a responsibility of fn `generate_simple_goal_activities`

moaz-mokhtar: Sorry for late response.

---
I did the change but it causing wrong scheduling for `budget_and_goal_one_day`
```shell
running 10 tests
test e2e::basic_1 ... ok
test e2e::filler_goal ... ok
test e2e::algorithm_challenge ... ok
test e2e::budget_and_goal_one_day ... FAILED
test e2e::bug_start_missing ... ok
test e2e::not_on ... ok
test e2e::after_12 ... ok
test e2e::sleep_1 ... ok
test e2e::budget_with_subgoal ... ok
test e2e::default_budgets ... ok

```

I have a challenge here.

I did it but it causing wrong scheduling although I have the same params.

I did investigate it and found below:
- When calling `adjust_parent_activities` from inside `generate_simple_goal_activities`, it causing wrong values for `calender_overlay` as below figure:
image

- When calling `adjust_parent_activities` from inside `lib.rs` (like current case), it causing accurate values which leads to accurate scheduling as below: 
image

---

So, I suggest to avoid investigate this issue at the time and solve more issues to achieve business requirements. Also I'm ready to continue investigating and resolving this issue based on your request.

tijlleenders: Please investigate what the issue is and propose a fix.  

I don't see why activities generated from budget goals should be affected by generating/adjusting activities from simple goals.

Ignoring the problem is not a solution.

moaz-mokhtar: I agree...
I will, thanks

tijlleenders:
The approach of adjusting Activities is convoluted and hard to follow. It' better to generate filler Activities as a separate step.

Placing in two separate steps is also not desired, as it implicitly says Budgets are less important than simple Goals - and they can't interact with each other to optimize placing.

PR #437, opened 2024-01-26
add not-on test case

tijlleenders:
The idea of this functionality is that one can reschedule/postpone Tasks for a certain period - or just mark some hours of a specific day as inconvenient for this particular goal.  

Frontend implementation is a bit simpler - just focusing on the postpone use case: https://github.com/tijlleenders/ZinZen/issues/1400


src/models/activity.rs
```diff
@@ -284,9 +296,19 @@ impl Activity {
     ) -> Vec<Activity> {
         let mut activities: Vec<Activity> = vec![];
 
-        let compatible_hours_overlay =
-            Self::get_compatible_overlay_hours(goal_to_use, calendar, time_budget);
-
+        let compatible_hours_overlay = Activity::get_compatible_hours_overlay(
```
thinkrapido, line 299: why is this code duplicated?

thinkrapido (commented):
Why is this code refactored into code duplication?

Tushar-4781 (approved):

src/models/activity.rs
tijlleenders: See https://github.com/tijlleenders/ZinZen-scheduler/pull/446

PR #439, opened 2024-01-26
Tijl/-/add-test-impossible

tijlleenders:
Any simple goal that has a deadline and is not possible before the deadline should be marked as impossible.

However - if the simple goal has no deadline it should never be marked as impossible.


tijlleenders:
@moaz-mokhtar After #434 you can try this one?

Issue #426, opened 2023-12-15
Format tests with on_days as budgets


tijlleenders:
**Is your feature request related to a problem? Please describe.**
Anything that has an on_day filter is a budget - since that feature is not available for 'simple' goals.

**Describe the solution you'd like**
Go through all tests in `tests/jsons/validated/` and `tests/jsons/stable.bak` and transform the input.json to a budget goal for any goal that has an on_day filter. 

**Describe alternatives you've considered**
Allow on_day on 'simple' goals. This goes back on frontend UX decision to split budgets from goals for easier input interface. Budgets are done once - or at least infrequently.  Goals can make use of a budget by being a sub-goal of a budget.

**Additional context**
An example of a `input.json` that needs to be transformed is `tests/jsons/stable.bak/before-7/input.json`

PR #440, opened 2024-01-28
Rom/ /refactor

thinkrapido:


tijlleenders (approved):
@thinkrapido Thanks for the great improvements!  
It was painful to have the `tests_mod.rs` duplicate the run_scheduler logic... this is much more ergonomic. 👍🏽

@thinkrapido @kobe-reygel  
The tests in `/stable.bak` and `/validated` are no longer in the right format and need to be reprocessed/adapted/checked again - to conform to the new approach of separating simple goals from budget goals - as well as the new input format. I've made #426 for that.
I'm sorry if that wasn't clear.

PR #446, opened 2024-02-03
undo duplicating function

### LINKED ISSUE #445: Resolve confusingly similarly named functions
**Is your feature request related to a problem? Please describe.**
image

**Describe the solution you'd like**
Rename/Refactor so that code reads well.  
In my opinion, slight code duplication is allowed if this makes it more readable.  

**Describe alternatives you've considered**
None.

**Additional context**
None.

tijlleenders:
Resolves #445

Refactoring out this code in this way results in a function that is just a wrapper for a nearly identical function - which is confusing.  
It is more readable to adjust the inputs inline - then call the generic function once.

### DIFF
--- a/src/models/activity.rs
+++ b/src/models/activity.rs
@@ -284,9 +284,18 @@ impl Activity {
     ) -> Vec<Activity> {
         let mut activities: Vec<Activity> = vec![];
 
-        let compatible_hours_overlay =
-            Self::get_compatible_overlay_hours(goal_to_use, calendar, time_budget);
-
+        let compatible_hours_overlay = Activity::get_compatible_hours_overlay(
+            calendar,
+            goal_to_use.filters.clone(),
+            calendar
+                .start_date_time
+                .sub(Duration::hours(24)) //TODO: fix magic number
+                .add(Duration::hours(time_budget.calendar_start_index as i64)),
+            calendar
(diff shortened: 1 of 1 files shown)


src/models/activity.rs
```diff
@@ -284,9 +284,18 @@ impl Activity {
     ) -> Vec<Activity> {
         let mut activities: Vec<Activity> = vec![];
 
-        let compatible_hours_overlay =
-            Self::get_compatible_overlay_hours(goal_to_use, calendar, time_budget);
-
+        let compatible_hours_overlay = Activity::get_compatible_hours_overlay(
```
thinkrapido, line 287: Please undo this code duplication.

thinkrapido (changes requested):

src/models/activity.rs
tijlleenders: Actually this is on purpose. See PR #445 for the explanation. I'm actually reducing code duplication. 

I agree more can be done ... but I'll leave that for later. It is not added in this PR. It was already there and can be handled in a separate issue - where we deduplicate at the right level, if this makes sense.

tijlleenders: There were two functions with nearly identical names before - I've deduplicated this to one.

PR #447, opened 2024-02-11
Rom/434/add filler goal test

thinkrapido:


tijlleenders:
Closing this in favor of similar implementation (based on looking at the parent), but one that modifies the Activities instead of the Goals.

